PR #15, opened 2016-11-21
[WIP] Fix Bug #10109

mrkn:
https://bugs.ruby-lang.org/issues/10109


mrkn:
I found that Bug #10109 has already been fixed.

PR #20, opened 2016-11-22
Brushing up documentation of `BigDecimal#abs`

yui-knk:
I tried to brushing up documentation (referring ruby/ruby documentation style).
How do you think about this?


mrkn:
The direction is OK.

But I don't merge this pull request because I'm going to modify the result of `BigDecimal#inspect` to be more friendly for resolving [[Feature #12161]](https://bugs.ruby-lang.org/issues/12161).  This modification will break the change in this pull request.

yui-knk:
Updated to more friendly inspect versions.

PR #48, opened 2016-12-18
Brushing up documentation of math.rb and util.rb

koic:
It seems that this behavior will change in Ruby 2.4.0+.

I used version of Ruby is `ruby 2.4.0dev (2016-12-18 trunk 57108) [x86_64-darwin13]`.

Thanks.

### DIFF
--- a/lib/bigdecimal/math.rb
+++ b/lib/bigdecimal/math.rb
@@ -26,7 +26,7 @@ require 'bigdecimal'
 #   include BigMath
 #
 #   a = BigDecimal((PI(100)/2).to_s)
-#   puts sin(a,100) # => 0.10000000000000000000......E1
+#   puts sin(a,100) # => 0.99999999999999999999......e0
 #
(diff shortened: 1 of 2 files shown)


lib/bigdecimal/math.rb
```diff
@@ -26,7 +26,7 @@
 #   include BigMath
 #
 #   a = BigDecimal((PI(100)/2).to_s)
-#   puts sin(a,100) # => 0.10000000000000000000......E1
+#   puts sin(a,100) # => 0.99999999999999999999......e0
```
koic: I'm not an expert. It seems that this results changed before Ruby 2.3.3.

mrkn: I found that it changed long time ago, at least before 1.1.0.

koic: Understood. Thanks for your information.

PR #49, opened 2016-12-21
Change `#round` to accept `{ half: nil }` to follow r57130

yui-knk:
By r57130 `Float#round` is changed to accept `{ half: nil }`.
This commit follows the change.
`{ half: nil }` means the default behavior, and the default behavior
of `BigDecimal#round` is to use `BigDecimal::ROUND_MODE`.


yui-knk:
```ruby
:bigdecimal kaneko$ ruby -v
ruby 2.4.0dev (2016-12-21 trunk 57133) [x86_64-darwin14]
:bigdecimal kaneko$ irb
irb(main):001:0> (1.1).round half: nil
=> 1
```

Issue #52, opened 2017-01-10
Can we drop ruby 1_9_3 from CI matrix?


yui-knk:
Tests for Ruby 1.9.3 have been failed.
There are some reasons why tests on 1.9.3 do not work.

* Invalid syntaxes about keyword args (e.g. `**opt` or `method(a: :b)`)
* A missed macoro (`NUM2USHORT `)

I tried to run tests on 1.9.3 (https://github.com/yui-knk/bigdecimal/commits/1_9_3), but I feel it is too complex. So I propose to drop 1.9.3 from CI 😢

mrkn:
It's good idea.

PR #54, opened 2017-01-16
Drop ruby 1_9_3 from CI matrix

yui-knk:
Tests for Ruby 1.9.3 have been failed.
There are some reasons why tests on 1.9.3 do not work.

* Invalid syntaxes about keyword args (e.g. `**opt` or `method(a: :b)`)
* A missed macoro (`NUM2USHORT `)

Changing codes to run tests on 1.9.3 is too complex.
So drop 1.9.3 from CI.

close https://github.com/ruby/bigdecimal/issues/52

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,7 @@ rvm:
 - 2.2.6
 - 2.1.10
 - 2.0.0
-- 1.9.3
 matrix:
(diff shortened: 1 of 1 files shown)


mrkn:
Thank you very much for your work.

Issue #51, opened 2016-12-22
BigDecimal regression - to_d inconsistent with other numeric classes


jhawthorn:
I originally filed this on [bugs.ruby-lang.org](https://bugs.ruby-lang.org/issues/13062), not sure if this is a more appropriate place.

As of ruby 2.4.0-rc1, `BigDecimal()` was changed to raise exceptions on invalid input, which is more consistent with the other numeric types. Unfortunately, `String#to_d` now also raises errors, which is inconsistent with the other `to_*` methods (`to_i`, `to_f`), which return 0 on error.

Under ruby 2.4.0-rc1:

``` ruby
> require 'bigdecimal'
> require 'bigdecimal/util'
> "invalid".to_d
ArgumentError: invalid value for BigDecimal(): "invalid"
> "invalid".to_i
=> 0
> "invalid".to_f
=> 0.0
```

Under ruby 2.3.3 or 2.4.0preview3:

``` ruby
> "invalid".to_d
=> #<BigDecimal:55871ca1f808,'0.0',9(9)>
> "invalid".to_i
=> 0
> "invalid".to_f
=> 0.0
```

There's also a further problem that BigDecimal() still doesn't behave the same as Integer() when given a string with the number at the start:

Under ruby 2.4.0-rc1:

``` ruby
> BigDecimal("2 turtle doves")
=> 0.2e1
> Integer("2 turtle doves")
ArgumentError: invalid value for Integer(): "2 turtle doves"
> Float("2 turtle doves")
ArgumentError: invalid value for Float(): "2 turtle doves"
```

So `BigDecimal` is still inconsistent.

mrkn:
The patch posted by Piotr Szmielew in Redmine is good to fix this.
I'll apply this patch later.
Thanks for reporting the bug.

By the way,

> I originally filed this on bugs.ruby-lang.org, not sure if this is a more appropriate place.

Both places are appropriate.
But I prefer to use GitHub issue because it is more useful than Redline.

esse:
@mrkn Should I make pull request with that? Or is patch file provided through redmine enough?

mrkn:
@esse Would you please submit your pull request?
I want to import your contribution as your commit.

esse:
@mrkn pull request added: https://github.com/ruby/bigdecimal/issues/55

mrkn:
@esse Thank you.  I'll check it at this weekend.

abezzub:
[...]

PR #55, opened 2017-01-16
fixes issue: #13062 - String#to_d is not raising error

esse:
This commit fixes issue with String#to_d method which was inconsistent
with other to_d methods (String one returned error, while other simply
    returns 0).

Also, there tests added for this case.

Associated GH issue: https://github.com/ruby/bigdecimal/issues/51

### DIFF
--- a/lib/bigdecimal/util.rb
+++ b/lib/bigdecimal/util.rb
@@ -58,7 +58,11 @@ class String
   #     # => 0.5e0
   #
   def to_d
-    BigDecimal(self)
+    begin
+      BigDecimal(self)
+    rescue ArgumentError
+      BigDecimal(0)
+    end
   end
 end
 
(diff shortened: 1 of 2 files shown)


mrkn:
Thanks a lot !!

pwnall:
@mrkn Can you please do a 1.3.1 release with this fix? Thank you very much!

mrkn:
@pwnall I've released 1.3.1.  Please check it.

pwnall:
@mrkn Thank you very much for your quick response! The new release works for me!

nzifnab:
Hmm, just chiming in here, but there's still an inconsistency with other number classes:

    >> "30.".to_d
    => 0.0
    >> "30.".to_f
    => 30.0

(A trailing decimal point is ignored in floats, but gives that `ArgumentError` and subsequently returns 0 for `BigDecimal`).

Should I make a new issue for this?

PR #57, opened 2017-02-07
Fix documentation errors for 'bigdecimal/util'

stomar:
Fix errors in the documentation of util.rb:

* bigdecimal/util extends the native classes with #to_d, not BigDecimal
* BigDecimal is extended with #to_d and #to_digits, not Numeric

_I already opened an issue on bugs.ruby-lang.org, [#13195](https://bugs.ruby-lang.org/issues/13195), but I guess the correct place is here?_


mrkn:
> I guess the correct place is here?

Yes it is. Thank you for your concern.

PR #61, opened 2017-02-12
Fix error message for Rational#to_d with invalid precision

stomar:
Make the error message for nonpositive precision in Rational#to_d
match with the error raised by BigDecimal.new in that case.

Current behavior:

``` ruby
require "bigdecimal"
require "bigdecimal/util"

r = 22/7r  # => (22/7)

BigDecimal(r, -1) rescue $!  # => #<ArgumentError: argument must be positive>
r.to_d(-1)        rescue $!  # => #<ArgumentError: negative precision>
```

Note also that the condition (`precision <= 0`) does not match the current error message.

### DIFF
--- a/lib/bigdecimal/util.rb
+++ b/lib/bigdecimal/util.rb
@@ -129,7 +129,7 @@ class Rational < Numeric
   #   # => 0.314e1
   def to_d(precision)
     if precision <= 0
-      raise ArgumentError, "negative precision"
+      raise ArgumentError, "argument must be positive"
     end
     num = self.numerator
     BigDecimal(num).div(self.denominator, precision)


mrkn:
I want to fix the error message of `BigDecimal` method instead of `to_d` because I think `negative precision` is more informative about the error than `argument must be positive`.

mrkn:
@stomar So please modify your patch if you could.

stomar:
What about precision 0? Current behavior:

``` ruby
r.to_d(0)        rescue $!  # => #<ArgumentError: negative precision>
```

which is confusing (or, more precisely, wrong).

Should to_d be changed to also accept a precision of 0?

stomar:
I suggest postponing the decision on the best wording for the error message until the desired behavior of Rational#to_d and BigDecimal() regarding zero precision is clear. See #64.

stomar:
Closed in favor of #65.

PR #65, opened 2017-02-16
Fix error message for precision argument

stomar:
Current behavior:

``` ruby
BigDecimal(-0.2, 0)              # => -0.200000000000000011102230246251565404e0
BigDecimal(-0.2, -1)  rescue $!  # => #<ArgumentError: argument must be positive>

d = (22/7r).to_d(9)

d.div(2, 3)              # => 0.157e1
d.div(2, 0)              # => 0.157142857e1
d.div(2, -1)  rescue $!  # => #<ArgumentError: argument must be positive>
```

The error message could be more descriptive (which argument?), furthermore, the message is wrong, since precisions can also be 0. So, something like e.g. "negative precision" would be better. Problem: the exception is raised by `GetPositiveInt()`, which is also used by BigDecimal#to_s:

``` ruby
d.to_s(3)                # => "0.314 285 714e1"
d.to_s(0)                # => "0.314285714e1"
d.to_s(-1)  rescue $!    # => #<ArgumentError: argument must be positive>
```
In this case, the argument is _not_ a precision. So, simply changing the message would not work (and the very general "negative argument" would be misleading for the cases with more than one numeric argument).

The PR separates argument checking for to_s from the other cases that use GetPositiveInt(), allowing for a different error message. It also disallows value 0 for BigDecimal#to_s, which doesn't make sense for grouping of digits, and which is simply ignored currently.

The new behavior would be

``` ruby
d.to_s(3)                # => "0.314 285 714e1"
d.to_s(0)   rescue $!    # => #<ArgumentError: argument must be positive>
d.to_s(-1)  rescue $!    # => #<ArgumentError: argument must be positive>

BigDecimal(-0.2, 0)              # => -0.200000000000000011102230246251565404e0
BigDecimal(-0.2, -1)  rescue $!  # => #<ArgumentError: negative precision>

d.div(2, 3)              # => 0.157e1
d.div(2, 0)              # => 0.157142857e1
d.div(2, -1)  rescue $!  # => #<ArgumentError: negative precision>
```

I verified that all other use cases for GetPositiveInt() are for precision (number of significant digits), namely ::new, div, add, sub, mult, sqrt.

I am looking forward to any feedback, and have no problems with modifying the PR if necessary (coding style, adding tests, etc.).
This is more or less my first "project" for learning about Ruby's C API, so thanks for your patience.

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -2029,6 +2029,7 @@ BigDecimal_to_s(int argc, VALUE *argv, VALUE self)
     char  *psz;
     char   ch;
     size_t nc, mc = 0;
+    SIGNED_VALUE m;
     VALUE  f;
 
     GUARD_OBJ(vp, GetVpValue(self, 1));
@@ -2059,7 +2060,11 @@ BigDecimal_to_s(int argc, VALUE *argv, VALUE self)
 	    }
 	}
 	else {
-	    mc = (size_t)GetPositiveInt(f);
+	    m = NUM2INT(f);
+	    if (m <= 0) {
+		rb_raise(rb_eArgError, "argument must be positive");
+	    }
+	    mc = (size_t)m;
 	}
     }
     if (fmt) {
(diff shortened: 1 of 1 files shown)


stomar:
@mrkn Any thoughts on this?

mrkn:
@stomar thank!!

PR #71, opened 2017-05-10
Update Travis for latest version Rubies

junaruga:
Hello,

I want to update Rubies in `.travis.yml` for testing.

Thanks.

PR #72, opened 2017-05-24
Add condition for mathn test. mathn was removed from Ruby 2.5

hsbt:
Resolve https://github.com/ruby/bigdecimal/pull/71#issuecomment-300696948

I added an ugly condition for Ruby 2.5.


mrkn:
R.I.P. mathn :pray:

PR #75, opened 2017-07-12
Added cross compilation task using rake-compiler-dock

hsbt:
Hi, I added `build:mingw` task for mingw platforms. It makes that we can provide precompiled gem for mingw platform. It helps to use the latest version of bigdecimal for Windows users without devkit.

@mrkn How do you think about this?

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,4 +1,6 @@
-require "bundler/gem_tasks"
+require "bundler"
+Bundler::GemHelper.install_tasks
+
 require "rake"
 require "rake/extensiontask"
 require "rake/testtask"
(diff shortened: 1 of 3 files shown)


mrkn:
It's very helpful for me because I don't have enough knowledge and experience on Windows environments.

Thank you very much for your works.

mrkn:
@hsbt how can I push pre-built gem files for mingw32 platform?

hsbt:
@mrkn You can push .gem files to rubygems.org directly for mignw32 platform.  Like this:

```
$ gem push pkg/psych-3.0.0.beta2-x64-mingw32.gem 
```

You can see https://rubygems.org/gems/psych/versions/3.0.0.beta2-x64-mingw32

In Mingw32/Windows environment, You can install it via rubygems.

Issue #84, opened 2017-12-13
[Feature-13984] BigDecimal should be immutable/frozen and return itself on #dup


mrkn:
https://bugs.ruby-lang.org/issues/13984

Currently, BigDecimal instances are not frozen, and supports #dup by returning a new instance.

In the spirit of [#12979](https://bugs.ruby-lang.org/issues/12979) and [#13983](https://bugs.ruby-lang.org/issues/13983), I think BigDecimal instances should always be frozen.
Then BigDecimal#dup/clone should just return the same instance.

mrkn:
This fixes by e089c3be29bc4314c583237ab35cba1528b33d62

PR #85, opened 2017-12-13
Deprecate BigDecimal#clone and #dup

mrkn:
For removing clone and dup instance methods in the future [#84],
we first deprecate them during Ruby 2.5 period.

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -2630,6 +2630,20 @@ BigDecimal_initialize_copy(VALUE self, VALUE other)
     return self;
 }
 
+static VALUE
+BigDecimal_clone(VALUE self)
+{
+  rb_warning("BigDecimal#clone is deprecated.");
+  return rb_call_super(0, NULL);
+}
+
+static VALUE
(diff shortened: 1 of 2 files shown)


eregon:
Sounds good, should this be part of 2.5 RC1 or it will be integrated later?

mrkn:
@eregon Yes.

matthewd:
This means we can't safely call `dup` on an unknown object (without getting a warning), just in case it's a BigDecimal.

That seems inconsistent with the goal of https://bugs.ruby-lang.org/issues/12979.

Please consider skipping the deprecation, and just silently switching `dup` to return self when you change to frozen-by-default. That's worked safely for a number of core types, so I think it would be fine for BigDecimal too.

mrkn:
@matthewd Removing BigDecimal#clone and #dup doesn't mean undefine them.
Numeric class, the super class of BigDecimal, has clone and dup, and they just return the receiver.

mrkn:
I misunderstood what you said in the previous comment.
But now I understood.  It's make sense, so I'll stop issuing deprecation warning and let them just return receivers.

Thank you.

PR #86, opened 2017-12-13
Deprecate BigDecimal.new

mrkn:
To remove `BigDecimal.new` from Ruby 2.6, this should be deprecated during Ruby 2.5 period.

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -392,7 +392,7 @@ BigDecimal_hash(VALUE self)
  *
  * Method used to provide marshalling support.
  *
- *      inf = BigDecimal.new('Infinity')
+ *      inf = BigDecimal('Infinity')
  *        #=> Infinity
  *      BigDecimal._load(inf._dump)
  *        #=> Infinity
@@ -875,7 +875,7 @@ BigDecimal_to_r(VALUE self)
  * be coerced into a BigDecimal value.
  *
  * e.g.
- *   a = BigDecimal.new("1.0")
+ *   a = BigDecimal("1.0")
  *   b = a / 2.0 #=> 0.5
  *
  * Note that coercing a String to a BigDecimal is not supported by default;
@@ -1164,7 +1164,7 @@ BigDecimal_comp(VALUE self, VALUE r)
  *
  * Values may be coerced to perform the comparison:
  *
- *   BigDecimal.new('1.0') == 1.0  #=> true
+ *   BigDecimal('1.0') == 1.0  #=> true
  */
 static VALUE
 BigDecimal_eq(VALUE self, VALUE r)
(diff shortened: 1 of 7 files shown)


rafaelfranca:
Do you have the ticket with the reasoning to remove this method? Is it to match `Interger()`?

mrkn:
The reason is to match core numeric classes like Integer, Float, and Rational.

I've created the issue #89 to remove BigDecimal.new at Ruby 2.6.

rafaelfranca:
Thanks! That make a lot of sense

baweaver:
This may be the case, but it causes a lot of downstream errors and is a breaking change done on a minor version. What would be the harm in leaving this syntax and encouraging the new syntax instead?

I feel we should be very careful in deprecations like this as the downstream impact is substantial. It creates inconsistencies in Ruby and causes a lot of confusion for newer engineers.

Consider that we have:

| `Kernel` method | Constructor  |
|-----------------|--------------|
| `Kernel#String` | `String.new` |
| `Kernel#Hash`   | `Hash.new`   |
| `Kernel#Array`  | `Array.new`  |

...and several others. Across Ruby this is very inconsistent, and is very confusing to me.

mrkn:
BigDecimal is a subclass of Numeric, and core Numeric subclasses such as Integer don’t have `new` class method.  So removing `BigDecimal.new` increases the consistency.

The Kernel methods you listed up are provided for conversion.  `new` methods are provided for just construction.  They have different roles.

baweaver:
I do not argue with the consistency, I argue with the deprecation in a minor version causing several downstream repositories to break.

The cost of leaving `new` in is substantially less than that of removing it for consistency.

mrkn:
You can select version 1.3.5 if you need `BigDecimal.new`.

ext/bigdecimal/bigdecimal.c
```diff
@@ -2594,6 +2594,13 @@ static Real *BigDecimal_new(int argc, VALUE *argv);
  * ArgumentError:: If the +initial+ is a Float or Rational, and the +digits+
  *                 value is omitted, this exception is raised.
  */
+static VALUE
+BigDecimal_s_new(int argc, VALUE *argv, VALUE self)
+{
+  rb_warning("BigDecimal.new is deprecated");
```
eregon: As a note, maybe this should have been `rb_warn`, as `rb_warning` only warns with `$VERBOSE=true` / `ruby -w`, not the default.

mrkn: bigdecimal 1.4.x always warns in `BigDecimal.new`.
See https://github.com/ruby/bigdecimal/blob/v1.4.x/lib/bigdecimal.rb#L10

eregon: Right, but in e.g. MRI 2.5.3, bigdecimal 1.3.4 is default and so:
```
$ ruby -rbigdecimal -e 'p BigDecimal.new "2"'
0.2e1
```
does not warn (without -v/-w).

PR #90, opened 2017-12-14
Tweak warning message of `BigDecimal.new`. Make the message to contain the correct constructor method name

pocke:
Ref: https://github.com/ruby/bigdecimal/commit/533737338db915b00dc7168c3602e4b462b23503

The warning says "BigDecimal.new is deprecated", but the message does not have the correct constructor method name. So I think users confuse when users see the message.
This change adds the correct method name, that is `Kernel.BigDecimal`.
Like https://github.com/ruby/bigdecimal/blob/a11b449ef6acb43c6b8e93fc57c1228d5163a6eb/ext/bigdecimal/bigdecimal.c#L152

What do you think?


mrkn:
Good catch.  Thank you!

PR #91, opened 2017-12-14
Remove unnecessary `require 'thread'`

znz:
ref https://github.com/ruby/ruby/commit/6064132c42ffe99e9a0d3a6ba80a7932c6f7903d


mrkn:
I imported the original commit from ruby-trunk at 3fcc44948af62fde7a3dee6eecb443fa03829784
Thank you for reporting.

PR #102, opened 2018-06-03
Remove unused `__have_macro__` alias from extconf

v-kolesnikov:
Hi, team! 👋 

I simple was exploring the source code and came across this one, it seems like unused.

the code was introduced here: https://github.com/ruby/bigdecimal/commit/02b605375c273dfae8c029588c3588b6ab85cab2#diff-92ed9076bbff06f570247479dabdeaaaR4
and then unused since: https://github.com/ruby/bigdecimal/commit/84da504a972c0c5f4692e1b39b93e98fdbfc9a1b#diff-92ed9076bbff06f570247479dabdeaaa

### DIFF
--- a/ext/bigdecimal/extconf.rb
+++ b/ext/bigdecimal/extconf.rb
@@ -17,8 +17,6 @@ bigdecimal_version =
 
 $defs << %Q[-DRUBY_BIGDECIMAL_VERSION=\\"#{bigdecimal_version}\\"]
 
-alias __have_macro__ have_macro
-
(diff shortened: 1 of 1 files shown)


v-kolesnikov:
The building on TravisCI is broken on master since https://travis-ci.org/ruby/bigdecimal/builds/365390502
We probably have to allow failures for unstable Ruby versions with `allow failures: ruby-head` (https://docs.travis-ci.com/user/customizing-the-build#Rows-that-are-Allowed-to-Fail)

PR #103, opened 2018-08-15
update BigDecimal#inspect documentation

robotdana:
The documentation for BigDecimal#inspect refers to the previous behaviour. I've updated the documentation to the current behaviour.


mrkn:
@robotdana Thank you very much!

Issue #63, opened 2017-02-12
Inconsistency between String#to_d and String#to_f


stomar:
String#to_d behaves differently from String#to_f for strings starting with a number that is not followed by whitespace:

``` ruby
require "bigdecimal"
require "bigdecimal/util"

"40.5degree".to_f  # => 40.5
"40.5degree".to_d  # => 0.0
```

For many other cases they behave in the same way, e.g.:

``` ruby
"40.5 degree".to_f  # => 40.5
"40.5 degree".to_d  # => 0.405e2

"foo".to_f  # => 0.0
"foo".to_d  # => 0.0

"40.5e-2".to_f  # => 0.405
"40.5e-2".to_d  # => 0.405e0
```

feist:
Also another (potentially serious) issue.  If you pass "1.", to_f & to_i accept it, but to_d throws an exception.

2.4.1 :001 > require 'bigdecimal'
2.4.1 :002 > require 'bigdecimal/util'
2.4.1 :003 > "1.".to_i
 => 1 
2.4.1 :004 > "1.".to_f
 => 1.0 
2.4.1 :005 > "1.".to_d
ArgumentError: invalid value for BigDecimal(): "1."

feist:
I'm seeing something even more worrying with bigdecimal-1.3.2

2.4.1 :001 > require 'bigdecimal'
 => true 
2.4.1 :002 > require 'bigdecimal/util'
 => true 
2.4.1 :006 > "1.".to_i
 => 1 
2.4.1 :007 > "1.".to_f
 => 1.0 
2.4.1 :008 > "1.".to_d
 => 0.0

nzifnab:
I'm also seeing this issue. Looks like it's somewhat related to this fix: https://github.com/ruby/bigdecimal/pull/55

In ruby 2.3.3:

    >> "30.".to_d
    => #<BigDecimal:55adc4a76918,'0.3E2',9(18)>

In ruby 2.4.1 (and updating bigdecimal to 1.3.2):

    >> "30.".to_d
    => 0.0

elfassy:
also
```
> "1.23other".to_d
=> 0.123e1
> "1.23degree".to_d
=> 0.0
```

kakivis:
I guess the problem it that:
[...]

PR #104, opened 2018-09-28
when parsing strings, allow a trailing decimal point

pdg137:
This fixes an unnecessary breaking change introduced in commit 3081a627cebdc1fc119425c7a9f009dbb6bec8e8.  Numbers with a trailing decimal point, such as "123." are unambiguous, standard in science and engineering, and we were able to parse them before, so it seems like we should be able to handle them. Strings with almost any other kinds of junk at the end are allowed, such as "123.0." or "123,", so it was surprising to find that this particular form of number was not supported.

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -4046,7 +4046,7 @@ VP_EXPORT Real *
 VpAlloc(size_t mx, const char *szVal)
 {
     const char *orig_szVal = szVal;
-    size_t i, ni, ipn, ipf, nf, ipe, ne, dot_seen, exp_seen, nalloc;
+    size_t i, ni, ipn, ipf, nf, ipe, ne, exp_seen, nalloc;
     char v, *psz;
     int  sign=1;
     Real *vp = NULL;
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -4141,12 +4141,10 @@ VpAlloc(size_t mx, const char *szVal)
     ipf = 0;
     ipe = 0;
     ne  = 0;
-    dot_seen = 0;
     exp_seen = 0;
     if (v) {
         /* other than digit nor \0 */
         if (szVal[i] == '.') {    /* xxx. */
-            dot_seen = 1;
             ++i;
             ipf = i;
             while ((v = szVal[i]) != 0) {    /* get fraction part. */
@@ -4177,7 +4175,7 @@ VpAlloc(size_t mx, const char *szVal)
                 break;
         }
     }
-    if (((ni == 0 || dot_seen) && nf == 0) || (exp_seen && ne == 0)) {
+    if ((ni == 0 && nf == 0) || (exp_seen && ne == 0)) {
 	VALUE str = rb_str_new2(orig_szVal);
 	rb_raise(rb_eArgError, "invalid value for BigDecimal(): \"%"PRIsVALUE"\"", str);
     }
--- a/test/test_bigdecimal.rb
+++ b/test/test_bigdecimal.rb
@@ -65,7 +65,7 @@ class TestBigDecimal < Test::Unit::TestCase
   def test_global_new_with_invalid_string
     [
       '', '.', 'e1', 'd1', '.e', '.d', '1.e', '1.d', '.1e', '.1d',
-      'invlaid value'
+      'invalid value'
     ].each do |invalid_string|
       assert_raise_with_message(ArgumentError, %Q[invalid value for BigDecimal(): "#{invalid_string}"]) do
         BigDecimal(invalid_string)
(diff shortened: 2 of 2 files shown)


ioquatix:
Thanks for picking this up. I'm not sure I'm able to merge this, but perhaps @nobu can take a look?

nobu:
It is intentional, as `Float("123.")`.

pdg137:
That's interesting that you point out the difference with `Float()`.  `BigDecimal()` actually behaves more like `String#to_f` since it allows almost any other characters to be appended. None of the documentation for any these functions suggests that a trailing decimal point would be disallowed.

In fact, this points to a more dangerous problem:

```
irb(main):011:0> "1.0e10".to_f
=> 10000000000.0
irb(main):012:0> "1.e10".to_f
=> 1.0
irb(main):026:0> BigDecimal("1.e10")
Traceback (most recent call last):
        3: from /opt/rubies/ruby-2.5.1/bin/irb:11:in `<main>'
        2: from (irb):26
        1: from (irb):26:in `BigDecimal'
ArgumentError (invalid value for BigDecimal(): "1.e10")
```

Would you want to change `to_f` to raise an exception for consistency?  I think we should allow the trailing decimal point in both cases, since there is no ambiguity about the intent.

mrkn:
@pdg137 `BigDecimal()` function should follow the spec of `Float()`, so if you find behaviors of `BigDecimal()` which are different from `Float()` are bugs of `BigDecimal()`.

`Object#to_d` is corresponding to `Object#to_f`, but it has parsing bugs #63 now unfortunately. Please wait or make a pull-request to fix it.

pdg137:
So is this basic difference a bug?

```ruby
BigDecimal('1x') # => 0.1e1
Float('1x') # fails
```

I also noticed this one:

```ruby
BigDecimal('1d1') # => 0.1e2
Float('1d1') # fails
```

mrkn:
@pdg137 Yes, it is.  I'll fix it.

DanGrayson:
> It is intentional, as `Float("123.")`.

@mrkn - why is an error on such input good behavior to intend?  Since it works works in Fortran and C, and has done so forever, it seems to be an easily repaired misfeature.

Why should `Float(".123")` work and `Float("123.")` not work?

mrkn:
@DanGrayson he current ruby doesn't accept `.123` as a floating-point number literal.  But, ruby accepted it until the version 1.6.8.  The `.123` form had been rejected at [this commit](https://github.com/ruby/ruby/commit/bb7085084e32ad15c51688bf55b841dad4f880a3).

I think it is the remains of this old behavior that Float() method accepts `.123`.

DanGrayson:
Hmm?  I tested it in version 2.3.7 to get the statement I made:
```
irb(main):002:0> RUBY_VERSION
RUBY_VERSION
=> "2.3.7"
irb(main):003:0> Float(".123")
Float(".123")
=> 0.123
irb(main):004:0> Float("123.")
Float("123.")
ArgumentError: invalid value for Float(): "123."
	from (irb):4:in `Float'
	from (irb):4
	from /usr/bin/irb:11:in `<main>'
```

DanGrayson:
Also this more recent version:
```
irb(main):001:0> RUBY_VERSION
RUBY_VERSION
=> "2.5.3"
irb(main):002:0> Float(".123")
Float(".123")
=> 0.123
irb(main):003:0> Float("123.")
Float("123.")
Traceback (most recent call last):
        3: from /usr/local/opt/ruby/bin/irb:11:in `<main>'
        2: from (irb):3
        1: from (irb):3:in `Float'
ArgumentError (invalid value for Float(): "123.")
```

DanGrayson:
At least this behavior is good:
```
irb(main):009:0> ".123".to_f
".123".to_f
=> 0.123
irb(main):010:0> "123.".to_f
"123.".to_f
=> 123.0
irb(main):011:0> RUBY_VERSION
RUBY_VERSION
=> "2.5.3"
```

mrkn:
> The current ruby doesn't accept .123 as a floating-point number literal.

I wrote "a floating-point number *literal*", but not an argument of `Float()` method.

```
$ RBENV_VERSION=2.3 ruby -ve '.123'
ruby 2.3.7p456 (2018-03-28 revision 63024) [x86_64-darwin17]
-e:1: no .<digit> floating literal anymore; put 0 before dot
-e:1: syntax error, unexpected '.'
-e:1: warning: possibly useless use of a literal in void context
```

mame:
The conversion functions like `Integer()` and `Float()` are stricter than `.to_i` and `.to_f`.  In principle, they accept only a string that is valid as a literal for the type.  `Float()` rejects `123.` because it not a valid float literal.
According to this principle, `Float()` should also reject `.123`.  I guess that this is because `.123` was a valid literal until Ruby 1.6.8.  I wonder if it is worth breaking the compatibility just for principle.

```
$ ./bin/ruby-1.6.8 -ve 'p .123'
ruby 1.6.8 (2002-12-24) [i686-linux]
0.123
```

If you want a flexible conversion, you may want to use `.to_i` or `.to_f`.  Note that they are somewhat dangerous because of their flexibility; for example, `"1.e10".to_f #=> 1.0`, as @pdg137 said.  `1.e10` is not a valid float literal, and `.to_f` extracts the prefix that is a valid numerical literal (in this case, "1" is a valid).

DanGrayson:
Thanks for the correction.  It's interesting that literals have reverse behavior, disallowing a dot at the other end:
```
irb(main):001:0> 123.
123.
irb(main):002:0> .123
.123
SyntaxError: (irb):2: no .<digit> floating literal anymore; put 0 before dot
(irb):2: syntax error, unexpected '.', expecting '('
	from /usr/bin/irb:11:in `<main>'
irb(main):003:0> RUBY_VERSION
RUBY_VERSION
=> "2.3.7"
```

PR #117, opened 2018-12-22
Mention exception: argument

zverok:
Hopefully can be merged before Ruby 2.6 release.


mrkn:
@zverok Thank you very much!!

PR #128, opened 2019-01-10
Add version and default gem badges

janlelis:
This shows the current gem version and link to the gem pages on rubygems.org and stdgems.org

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 [![Build Status](https://travis-ci.org/ruby/bigdecimal.svg?branch=master)](https://travis-ci.org/ruby/bigdecimal)
 [![Build status](https://ci.appveyor.com/api/projects/status/4j596bhldmmtcjpf/branch/master?svg=true)](https://ci.appveyor.com/project/ruby/bigdecimal/branch/master)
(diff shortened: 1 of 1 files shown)


mrkn:
@janlelis Adding the link to rubygems.org is OK.  On the other hand, I don't know about stdgems.org.  What is this site?  While I saw [the page for bigdecimal in the site](https://stdgems.org/bigdecimal/), I couldn't get the important information for the development of bigdecimal there.  I want to understand the necessity of adding the link to stdgems.org.  Would you please explain it?

janlelis:
@mrkn The site's main purpose is to collect which default gems versions are bundled exactly with each version of Ruby itself.

Here is an example scenario where this info could be useful: If Ruby 1.2.3 would come with bigdecimal 0.1.2, but this version of bigdecimal had a security issue which would then later be fixed with bigdecimal 0.1.3, one could check if there version of Ruby (in this case 1.2.3) is affected.

The page should also serves as a general documentation of Ruby's standard library and make it easier to discover its contents and where to look further (for example, direct RDoc links, who is the maintainer, or the information where to submit bugs). If you have ideas for more helpful data to put there, I'd be happy to hear.

As for the badge, I think it could help people who find the repository, but do not know (yet) that the gem is part of Ruby's standard library and that they can just use it without having to install it first.

Issue #131, opened 2019-04-18
String#to_d does not return partially parsed results for trailing decimal


ibrahima:
Hi! We've been using Ruby 2.3 for a while and are trying to upgrade to Ruby 2.5. One issue that I'm noticing is that, perhaps because of the fix to #51 (in #55), the behavior of `String#to_d` is different than it used to be when the string starts with a number but is not a complete number. The behavior is now inconsistent with the behavior of `String#to_f`

I.e.

```ruby
require 'bigdecimal'
require 'bigdecimal/util'

'1.'.to_d
=> 0.0 # Returns 0.0, but used to return 1.0
'1.'.to_f
=> 1.0
```
The documentation still says:

> initial
>    The initial value, as an Integer, a Float, a Rational, a BigDecimal, or a String.
>    If it is a String, spaces are ignored and unrecognized characters terminate the value.

This works okay if I pass in things like "1 pizza" etc, but I guess this specific case of a number with a decimal and nothing after the decimal no longer works, which is kind of unfortunate for my specific use case.

```ruby
'1.'.to_d == '1.'.to_f
# returns true on 2.3, false on 2.5
```

I'm on Ruby 2.5.5 with BigDecimal 1.3.4.

Thanks!

ibrahima:
Ohhh, I didn't read the other issues correctly - this is a dupe of https://github.com/ruby/bigdecimal/issues/63

ibrahima:
Actually, still confused. I installed BigDecimal v1.4.3 via bundler but in my Rails app, I still am getting the wrong value for `"1.".to_d`

```
2.5.5 (rails)[1] » '1.'.to_d
=> 0.0
```

mrkn:
This was resolved by 1.4.4.

PR #132, opened 2019-04-18
String#to_d should match String#to_f behavior with trailing decimal or 'e'

ibrahima:
In `String#to_d`, it should be more forgiving for missing fractional or exponential parts, just like `String#to_f`. This adds specs and updates the behavior so that this is the case.

```ruby
$ irb
irb(main):001:0> require 'bigdecimal/util'
=> true
irb(main):002:0> '1.'.to_f
=> 1.0
irb(main):003:0> '1e'.to_f
=> 1.0
irb(main):004:0> '1.'.to_d
=> 0.0
irb(main):005:0> '1e'.to_d
=> 0.0
```
See #131 and #63 (in particular, https://github.com/ruby/bigdecimal/issues/63#issuecomment-297544464 mentions this issue). The fix in #87 did not fix this case.

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -4281,7 +4281,7 @@ VpAlloc(size_t mx, const char *szVal, int strict_p, int exc)
 
     psz[i] = '\0';
 
-    if (((ni == 0 || dot_seen) && nf == 0) || (exp_seen && ne == 0)) {
+    if (strict_p && (((ni == 0 || dot_seen) && nf == 0) || (exp_seen && ne == 0))) {
         VALUE str;
       invalid_value:
         if (!strict_p) {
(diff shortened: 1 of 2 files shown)


ibrahima:
I also wonder if this behavior of `String#to_f` should be documented in https://github.com/ruby/spec/blob/master/core/string/to_f_spec.rb because it's not totally clear that it does this, though the doc string implies so because it says:

> Extraneous characters past the end of a valid number are ignored.

And in this case, an extra '.' or 'e' after a valid number are simply extraneous characters.

mrkn:
@ibrahima Thank you very much.

ibrahima:
That's great, thank you for merging! Will it be possible to also merge this into the 1.4.x branch for a new 1.4.x release? I didn't realize that the master branch was for 2.0. Should I open another pull request? Thanks!

PR #134, opened 2019-04-25
Backport String#to_d fix to 1.4 branch

ibrahima:
Cherry-picking these commits from #132 back to the 1.4 branch. Not sure if there was a better way to do this.

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -4294,7 +4294,7 @@ VpAlloc(size_t mx, const char *szVal, int strict_p, int exc)
 
     psz[i] = '\0';
 
-    if (((ni == 0 || dot_seen) && nf == 0) || (exp_seen && ne == 0)) {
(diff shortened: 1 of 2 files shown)


ibrahima:
By the way, I have been trying to use my branch of BigDecimal in my local Gemfile to fix the issue in my application, but it seems like somehow I can't install bigdecimal from git (using a line like `gem 'bigdecimal', github: 'ibrahima/bigdecimal', branch: 'v1.4.x_with_fix'`) because somehow the gemspec file is different from what it looks like in the repository. It looks like it is somehow postprocessed:

```ruby
# -*- encoding: utf-8 -*-
# stub: bigdecimal 1.4.4 ruby lib
# stub: ext/bigdecimal/extconf.rbext/bigdecimal/util/extconf.rb

Gem::Specification.new do
|s|
  s.name = "bigdecimal".freeze
  s.version = "1.4.4"
```

and this causes the extconf.rb script to fail because it can't find the `bigdecimal_version="1.4.4"` line.

Any ideas on how to install this gem from GitHub? Thanks!

mrkn:
@ibrahima I'm sorry I don't know why the reason and how to resolve.

PR #135, opened 2019-04-25
Update docs on `BigDecimal#round`

mrcasals:
When calling `BigDecimal#round` without any argument, it casts the result to an `Integer`, while calling it with an argument (`BigDecimal#round(n)`) keeps the result as a `BigDecimal`.

I think this is an unexpected behavior, but changing it might not be backwards-compatible, so let's document it at the docs at least.

This is a port of https://github.com/ruby/ruby/pull/2149 (in case the correct place to file the PR is here instead of there).

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -1756,11 +1756,15 @@ BigDecimal_fix(VALUE self)
  * round(n, mode)
  *
  * Round to the nearest integer (by default), returning the result as a
- * BigDecimal.
+ * BigDecimal if n is specified, or as Integer if it isn't.
+ *
  *
  *	BigDecimal('3.14159').round #=> 3
  *	BigDecimal('8.7').round #=> 9
  *	BigDecimal('-9.9').round #=> -10
+ *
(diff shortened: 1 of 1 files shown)


ext/bigdecimal/bigdecimal.c
```diff
@@ -1756,12 +1756,15 @@ BigDecimal_fix(VALUE self)
  * round(n, mode)
  *
  * Round to the nearest integer (by default), returning the result as a
- * BigDecimal.
+ * BigDecimal if n is specified, or as Integer if it isn't.
```
mrkn: `as an Integer`?

mrkn:
Thank you for your work.  Please fix for my comment.

> I think this is an unexpected behavior,

No, it is expected behavior. This behavior follows Float#round's behavior.

ext/bigdecimal/bigdecimal.c
mrcasals: Fixed!

mrkn: Thank you very much!

mrcasals:
Yay, thanks for merging this!

PR #138, opened 2019-05-08
Let util.so explicitly depends on bigdecimal.so

mrkn:


mrkn:
Windows support is broken: https://ci.appveyor.com/project/ruby/bigdecimal/builds/24385035/job/2imiqifrgcl5mal9

PR #145, opened 2019-05-10
Make BigDecimal::VERSION to support -dev suffix

mrkn:


nobu:
This doesn't work when only bigdecimal.gemspec is changed, since it does not trigger extconf.rb.

PR #148, opened 2019-07-14
Speed up loading bigdecimal gem

amatsuda:
This patch drastically reduces time taken to load the gem in all non-Windows platforms.

Here's a quick benchmark result of requiring the gem before and after this patch on my Mac.

```
ruby -e "now = Time.now; gem 'bigdecimal', '1.4.4'; require 'bigdecimal'; p '1.4.4' => Time.now - now"
ruby -e "now = Time.now; gem 'bigdecimal', '2.0.0.dev'; require 'bigdecimal'; p '2.0.0.dev' => Time.now - now"
{"1.4.4"=>0.172034}
{"2.0.0.dev"=>0.003902}
```

This patch was originally written by @nobu for io-console gem, and I'm just copying the solution to this repo, so please read other repos' PRs for more details.
https://github.com/ruby/psych/pull/406
https://github.com/ruby/io-console/pull/4

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,6 +6,7 @@ require "rake/extensiontask"
 require "rake/testtask"
 
 spec = eval(File.read('bigdecimal.gemspec'))
+spec.require_paths.insert(0, *%w[stub])
 Rake::ExtensionTask.new('bigdecimal', spec) do |ext|
   ext.lib_dir = File.join(*['lib', ENV['FAT_DIR']].compact)
   ext.cross_compile = true
--- a/bigdecimal.gemspec
+++ b/bigdecimal.gemspec
@@ -19,7 +19,6 @@ Gem::Specification.new do |s|
     bigdecimal.gemspec
     ext/bigdecimal/bigdecimal.c
     ext/bigdecimal/bigdecimal.h
-    lib/bigdecimal.rb
     lib/bigdecimal/jacobian.rb
     lib/bigdecimal/ludcmp.rb
     lib/bigdecimal/math.rb
@@ -28,6 +27,7 @@ Gem::Specification.new do |s|
     sample/linear.rb
     sample/nlsolve.rb
     sample/pi.rb
+    stub/bigdecimal.rb
   ]
 
   s.required_ruby_version = Gem::Requirement.new(">= 2.3.0".freeze)
--- a/lib/bigdecimal.rb
+++ b/lib/bigdecimal.rb
@@ -1,5 +1,0 @@
-begin
-  require "#{RUBY_VERSION[/\d+\.\d+/]}/bigdecimal.so"
-rescue LoadError
-  require 'bigdecimal.so'
-end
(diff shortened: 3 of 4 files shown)


kou:
I think that we can just stop fat gem support for the performance improvement.
RubyInstaller provides Ruby that can build an extension with `gem install bigdecimal`. Because RubyInstaller provides Ruby with MinGW as Devkit.

nobu:
Isn't a binary gem also for mswin?
Although mswin binary package may not exist now.

kou:
> Isn't a binary gem also for mswin?

We may be able to use a binary gem for \*-mingw32 with \*-mswin\* because it for \*-mingw32 uses non versioned `msvcrt.dll`. But it's not recommended. \*-mswin\* users should build their extension library by Visual C++.

And existing binary gems for \*-mingw32 aren't installed automatically with \*-mswin\* Ruby. Because they are different architecture.

amatsuda:
> I think that we can just stop fat gem support for the performance improvement.

I'm totally fine with either solution.

> RubyInstaller provides Ruby that can build an extension with gem install bigdecimal. Because RubyInstaller provides Ruby with MinGW as Devkit.

This may work great so far as everything is working. But it may cause some compilation troubles especially for those who are not very much familiar with compiling things, and so the development team may have to spend some time on shooting such troubles.

kou:
> But it may cause some compilation troubles especially for those who are not very much familiar with compiling things, and so the development team may have to spend some time on shooting such troubles.

Providing fat gem spends more time than resolving build error problems in my experience (I stopped fat gem support for Ruby-GNOME2 that has Windows users).

Normally, build error problems aren't occurred while AppVeyor build is green.

Normally, releasing fat gems is forgotten. For example, bigdecimal started fat gem support since 2017 #75 but no fat gems are released: https://rubygems.org/gems/bigdecimal/versions
The latest io-console (0.4.8) is the same situation: https://rubygems.org/gems/io-console/versions

I'm not surprised at this because releasing fat gems is a bother. We need to run more commands (`rake build:mingw` and `gem push pkg/*-mingw32.gem`) than `rake release` and prepare Docker. And we need to update supported Ruby version list when new Ruby is released: https://github.com/ruby/bigdecimal/blob/master/Rakefile#L21

I think that removing fat gem support will make both users (improving `require` performance) and developers (less maintenance cost) happy.

amatsuda:
> The latest io-console (0.4.8) is the same situation: https://rubygems.org/gems/io-console/versions

Ugh. So sorry about this. I asked nobu to release the gem and I was physically pairing with him, but yeah, we skipped that process...

Anyway, I'm now +1 for omitting the fat gem release. Let's see how it goes, and we can switch back to the fat gems anytime if we're seeing unexpected problems.
I suppose it's actually a very good timing to change the distribution process since the next release is a major version bump, right?

kou:
We don't need to care about major version bump. Because we have never released fat gems for bigdecimal.

@mrkn What do you think about removing fat gem support? I think that you don't want to care about Windows. So removing fat gem support seems reasonable to you.

mrkn:
I want to drop fat gem support.

kou:
OK. Let's drop fat gem support.

@amatsuda Do you want to work on it in this pull request or create another pull request?

amatsuda:
Just made another PR #149 for dropping the whole "fat gem" thing.

PR #150, opened 2019-08-15
Do not mutate frozen BigDecimal argument in BigMath.exp

jeremyevans:
BigMath.exp would previously mutate a BigDecimal argument if the
argument was negative.  This makes a internal copy of the argument
to avoid the mutation.

Fixes Ruby Bug 8542


mrkn:
@jeremyevans Good catch, thank you very much!

PR #151, opened 2019-09-26
Update documentation for Kernel#BigDecimal() digits argument [ci skip]

jeremyevans:
I expected this to affect the significant digits if providing
the first argument as a String, but it does not.  I'm positive this
argument is ignored if the first argument is an Integer, but I'm
not positive that it is ignored if the first argument is a String,
as the value is passed to VpAlloc. However, if it has an effect,
I'm not sure what it is.

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -2681,9 +2681,8 @@ VpNewVarArg(int argc, VALUE *argv)
  *           If it is a String, spaces are ignored and unrecognized characters
  *           terminate the value.
  *
- * digits:: The number of significant digits, as an Integer. If omitted or 0,
- *          the number of significant digits is determined from the initial
- *          value.
+ * digits:: The number of significant digits, as an Integer. This is required if
+ *          +initial+ is a Float or Rational, and ignored otherwise.
  *
  *          The actual number of significant digits used in computation is
  *          usually larger than the specified number.


ext/bigdecimal/bigdecimal.c
```diff
@@ -2681,9 +2681,8 @@ VpNewVarArg(int argc, VALUE *argv)
  *           If it is a String, spaces are ignored and unrecognized characters
  *           terminate the value.
  *
- * digits:: The number of significant digits, as an Integer. If omitted or 0,
- *          the number of significant digits is determined from the initial
- *          value.
+ * digits:: The number of significant digits, as an Integer. This is required if
+ *          +initial+ is a Float or Rational, and ignored otherwise.
```
mrkn: `0` for `digits` has the special meaning that the number of significant digits is automatically determined from the given initial value.  For a Rational initial value, for example, the resulting BigDecimal is determined on the value of `BigDecimal.limit` and the behavior of `BigDecimal#div`.

jeremyevans: OK.  I wonder if there is a better way to word it.  I want the documentation to explain that it does not have an effect in cases where you would expect it would:

```ruby
BigDecimal(1.1234, 2)
# => 0.11e1
BigDecimal(Rational(11234,10000), 2)
# => 0.11e1
BigDecimal('1.1234', 2)
# => 0.11234e1

BigDecimal(11234.0, 2)
# => 0.11e5
BigDecimal(11234, 2)
# => 0.11234e5
```

PR #153, opened 2019-10-08
Undef BigDecimal#initialize_copy

### LINKED ISSUE #147: BigDecimal can be changed with initialize_copy, even though it's frozen
```
$ ruby -v -rbigdecimal -e 'bd=BigDecimal("4"); p bd; p bd.frozen?; bd.send :initialize_copy, BigDecimal("7"); p bd'
ruby 2.7.0dev (2019-06-27T13:32:28Z master 310a42e89c) [x86_64-linux]
0.4e1
true
0.7e1
```

Probably `BigDecimal#initialize_copy` should raise, much like:
```
> 2.3.send :initialize_copy, 2.4
FrozenError: can't modify frozen Float: 2.3
```

jeremyevans:
Both BigDecimal#clone and BigDecimal#dup return self, there is no
reason to have initialize_copy exposed as a Ruby method.

The same is true for initialize_clone and initialize_dup.

Fixes #147

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -3453,7 +3453,9 @@ Init_bigdecimal(void)
 
 
     /* instance methods */
-    rb_define_method(rb_cBigDecimal, "initialize_copy", BigDecimal_initialize_copy, 1);
+    rb_undef_method(rb_cBigDecimal, "initialize_copy");
+    rb_undef_method(rb_cBigDecimal, "initialize_clone");
+    rb_undef_method(rb_cBigDecimal, "initialize_dup");
     rb_define_method(rb_cBigDecimal, "precs", BigDecimal_prec, 0);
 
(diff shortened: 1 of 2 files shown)


eregon:
Thank you for the PR!

I wonder, it seems Integer does neither define nor undefine those `initialize_*` methods, it just inherits `Kernel#initialize_copy`, which raises FrozenError when called.
Do you think the same strategy could be used for BigDecimal?

jeremyevans:
@eregon I'm sure it could.  Here's a branch if you prefer that approach: https://github.com/jeremyevans/bigdecimal/tree/remove-initialize_copy

Personally, having private methods defined that always raise FrozenError seems unnecessary, but that approach does result in fewer method table entries, and I guess it is more consistent with core classes.

eregon:
@jeremyevans I think that approach is better because it's consistent with core classes (and if more `initialize_*` methods were added it'd still work). Could you make a PR from that branch? It looks great to me.

Thank you for making those PRs!
I wish I had time to make them myself but I'm a bit too busy recently.

PR #156, opened 2019-10-09
Remove definition of BigDecimal#initialize_copy

jeremyevans:
This leaves the default definition, which will raise FrozenError.

@eregon prefers this approach.  It's more similar to core classes, and results in fewer method entries.


mrkn:
@jeremyevans Thank you very much!

PR #157, opened 2019-10-18
Remove taint checking

jeremyevans:
This removes the taint checking.  Taint support is deprecated in
Ruby 2.7 and has no effect.  I don't think removing the taint
checks in earlier ruby versions will cause any problems.


mrkn:
@jeremyevans Thank you very much!

PR #171, opened 2020-11-24
Use a higher default precision for BigDecimal#power and #**

jeremyevans:
When a fractional power is given, increase the precision if the
precision isn't specified via power's second argument:

Float: increase by 15 (rough number of decimal precision in float)
BigDecimal: increase by adding similar precision modifier as done to
            calculate the base precision.
Rational: double the precision, since a BigDecimal is created, but
          the created BigDecimal uses the same precision.

Increasing the precision for these power calculations has the obvious
tradeoff of making the calculations slower.

Fixes Ruby Bug #17264


mrkn:
I ignore the results of macOS workers because these workers are unstable this week.

PR #186, opened 2021-03-02
Use test-unit gem instead of test-suite of ruby core

hsbt:
I removed test-suite from ruby/ruby repository and replaced test-unit gem and the minimum requirements of test helper.


test/bigdecimal/test_ractor.rb
```diff
@@ -6,7 +6,7 @@ class TestBigDecimalRactor < Test::Unit::TestCase
 
   def setup
     super
-    skip unless defined? Ractor
+    pend unless defined? Ractor
```
mrkn, line 9: I prefer `omit` to `pend` in this case.
```
omit_unless(defined? Ractor)
```

bigdecimal.gemspec
```diff
@@ -37,11 +37,4 @@ Gem::Specification.new do |s|
   ]
 
   s.required_ruby_version = Gem::Requirement.new(">= 2.4.0")
-
-  s.add_development_dependency "benchmark_driver"
-  s.add_development_dependency "fiddle"
-  s.add_development_dependency "rake", ">= 12.3.3"
-  s.add_development_dependency "rake-compiler", ">= 0.9"
-  s.add_development_dependency "minitest", "< 5.0.0"
-  s.add_development_dependency "irb"
```
mrkn, line 46: Why these parts are moved to Gemfile?

hsbt: Bundler already removed these part with new gem template. Because `add_development_dependency` is less feature than Gemfile like `git`, `group`, `ref` or other specifying function.

test/bigdecimal/test_ractor.rb
hsbt: Fixed.

hsbt:
@mrkn How about this?

bigdecimal.gemspec
mrkn: OK, I understood the current situation of the dependency specification.  Thanks.

mrkn:
@hsbt Thanks a lot!

PR #190, opened 2021-04-01
gemspec: Do not distribute the gemspec as a "file"

olleolleolle:
This PR removes 1 file from the list of distributed files.

Question: I guess this gem is standard-included with Ruby, so LICENSE.txt is probably not necessary to add. Nor README.md. Do you agree?

### DIFF
--- a/bigdecimal.gemspec
+++ b/bigdecimal.gemspec
@@ -16,7 +16,6 @@ Gem::Specification.new do |s|
   s.require_paths = %w[lib]
   s.extensions    = %w[ext/bigdecimal/extconf.rb]
   s.files         = %w[
-    bigdecimal.gemspec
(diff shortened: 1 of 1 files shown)


mrkn:
bigdecimal is also distributed as a normal gem library.  So, If my understanding is correct, gemspec file should be included in the package file.

olleolleolle:
My understanding is this:

- When you build a gem, the gemspec is converted into _metadata_.
- The `.gemspec` files are not essential to include `files` directive.

Example to support this belief:

The gem https://rubygems.org/gems/activestream has 0 files in its `files` directive. Here is what is available about it: https://my.diffend.io/gems/activestream/0.0.0

I hope that this clarifies why I made the Pull Request. 👋

PR #193, opened 2021-05-21
Include Ruby 2.7 in versions table

cllns:
When [this table was first written](https://github.com/ruby/bigdecimal/commit/8b76f25a86abef6d21284aba2db6c2ad4759f90c) (2019-01-10), Ruby 2.7 wasn't released yet (2019-12-25).

I was just looking at it to reference which versions work, and I was worried Ruby 2.7 might not work, but it seems to work the same as 2.6. (I confirmed there's a subclass warning still too, though of course there is).


olleolleolle (approved):
This clarifies the documentation.

cllns:
@mrkn Can you merge this? Just came across the supported versions table again and remembered I opened this PR last year to fix it

mrkn:
@cllns Thank you for reminding me.

Issue #199, opened 2021-11-09
bignum too big to convert into `long' (RangeError)


casperisfine:
On current ruby 3.1:
```
>> BigDecimal::VERSION
=> "3.1.0-dev"
>> BigDecimal(-10000000000000000000)
(irb):2:in `BigDecimal': bignum too big to convert into `long' (RangeError)
```

On 3.0:

```
>> BigDecimal::VERSION
=> "3.0.0"
>> BigDecimal(-10000000000000000000)
=> -0.1e20
```

casperisfine:
Actually, it seems that the trigger is Ruby 3.1 and not the latest bigdecimal version. I'll try to track this down.

casperisfine:
Ok, so the problem is in https://github.com/ruby/bigdecimal/commit/4792a917d806ca1059c952f489413073ea51bf01

When dealing with signed integers, comparing `rb_absint_size` to `sizeof long` is no longer correct.

cc @mrkn

mrkn:
@casperisfine Thank you for reporting and investigating this!

casperisfine:
Thanks for the merge!

PR #200, opened 2021-11-09
Fix negative Bignum conversion

casperisfine:
Fix: https://github.com/ruby/bigdecimal/issues/199

Introduced in 4792a917d806ca1059c952f489413073ea51bf01

`rb_absint_size` return the number of bytes needed to fit the absolute integer, but negative integers need the sign, so one more bit, and potentially one more byte.

I would have expected the same bug to happen between `LONG_MIN` and `-ULONG_MAX` but somehow I couldn't reproduce it.

### DIFF
--- a/test/bigdecimal/test_bigdecimal.rb
+++ b/test/bigdecimal/test_bigdecimal.rb
@@ -12,7 +12,12 @@ class TestBigDecimal < Test::Unit::TestCase
     require 'fiddle'
     LONG_MAX = (1 << (Fiddle::SIZEOF_LONG*8 - 1)) - 1
     LONG_MIN = [LONG_MAX + 1].pack("L!").unpack("l!")[0]
+    LLONG_MAX = (1 << (Fiddle::SIZEOF_LONG_LONG*8 - 1)) - 1
+    LLONG_MIN = [LLONG_MAX + 1].pack("Q!").unpack("q!")[0]
+    ULLONG_MAX = (1 << Fiddle::SIZEOF_LONG_LONG*8) - 1
     LIMITS = {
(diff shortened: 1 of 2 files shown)


casperisfine:
I modified the PR a bit, turns out `rb_absint_size()` second parameter gives us the number of leading zero bits. So we can know precisely wether we have enough space for a signed number.

test/bigdecimal/test_bigdecimal.rb
```diff
@@ -2089,6 +2089,16 @@ def test_initialize_copy_dup_clone_frozen_error
...
 
+  def test_llong_min
+    # https://github.com/ruby/bigdecimal/issues/199
+    # Between LLONG_MIN and -ULLONG_MAX
+    llong_min = -(2 ** 63 + 1)
+    assert_equal BigDecimal(llong_min.to_s), BigDecimal(llong_min)
+
+    minus_ullong_max = -(2 ** 64 - 1)
+    assert_equal BigDecimal(minus_ullong_max.to_s), BigDecimal(minus_ullong_max)
+  end
```
mrkn, line 2100:
```suggestion
  def test_llong_min_gh_200
    # https://github.com/ruby/bigdecimal/issues/199
    # Between LLONG_MIN and -ULLONG_MAX
    llong_min = -(2 ** 63 + 1)
    assert_equal(BigDecimal(llong_min.to_s), BigDecimal(llong_min), "[GH-200]")

    minus_ullong_max = -(2 ** 64 - 1)
    assert_equal(BigDecimal(minus_ullong_max.to_s), BigDecimal(minus_ullong_max), "[GH-200]")
  end
```

PR #201, opened 2021-11-09
CI: Quote 3.0 to avoid YAML Float-to-String issue

olleolleolle:
This helps the output of the "name" of the CI Job in the PR checks list not to say "3" when it meant "3.0".

This is a very small change, but which avoids a Float-to-String loss of characters.

An example (not from this project) of what it can look like before this change, in this picture:

bild

Read more details, if you like: https://github.com/actions/runner/issues/849


hsbt:
👍

PR #204, opened 2021-11-19
Enhanced RDoc for selected methods

BurdetteLamar:
Treated:
- #+@
- #add
- #+

A couple of things I've made consistent with usage over at ruby/ruby:

- The second argument to `add`, I've called `ndigits`.
- I've placed the RDoc for the methods immediately before the code, so that `Document-method` is not needed.

I also discovered that `+` and `add` are different, and have documented them separately.

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -1096,12 +1096,14 @@ BigDecimal_coerce(VALUE self, VALUE other)
 }
 
 /*
- * call-seq:
- *    +big_decimal  ->  big_decimal
+ *  call-seq:
+ *    +big_decimal -> big_decimal
  *
- * Return self.
+ *  Returns +self+:
  *
- *     +BigDecimal('5')  #=> 0.5e1
+ *     +BigDecimal(5)  # => 0.5e1
+ *     +BigDecimal(-5) # => -0.5e1
+ *
  */
 
 static VALUE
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
@mrkn, I don't want to put up anything more until you've had a chance to comment on this PR.  That will give me good guidance going forward.

ext/bigdecimal/bigdecimal.c
```diff
@@ -1849,6 +1845,31 @@ BigDecimal_div3(int argc, VALUE *argv, VALUE self)
...
+  *  Returns the sum of +self+ and +value+
+  *  with a precision of +ndigits+ decimal digits.
+  *
+  *  When +ndigits+ is less than the number of significant digits
+  *  in the sum, the sum is rounded to that number of digits,
+  *  according to the current rounding mode; see BigDecimal.mode.
+  *
+  *  Examples:
+  *
+  *    BigDecimal('111111.111').add(1, 0) # => 0.111112111e6
```
mrkn, line 1861: @BurdetteLamar Could you please add the line to specify the rounding mode explicitly before this line?

BurdetteLamar: Done.

BurdetteLamar:
@mrkn, is this ready to merge?

ext/bigdecimal/bigdecimal.c
mrkn: @BurdetteLamar Could you please use `:up` instead of `BigDecimal::ROUND_HALF_UP` because I want to promote using a symbol instead of a constant?

mrkn:
Thank you very much.

ext/bigdecimal/bigdecimal.c
BurdetteLamar: Done.  But I have used `:half_up` (the default) because that's how the examples were made.

By the way, I strongly agree that the symbol is better.

PR #211, opened 2021-12-02
Improve extconf to allow using bigdecimal as a git gem

casperisfine:
e.g.
```
gem "bigdecimal", github: "ruby/bigdecimal"
```

It would fail because bundler regenerates the `gemspec`, so `bigdecimal_version` is gone.

@mrkn

### DIFF
--- a/bigdecimal.gemspec
+++ b/bigdecimal.gemspec
@@ -1,10 +1,8 @@
 # coding: utf-8
 
-bigdecimal_version = '3.1.0.dev'
-
 Gem::Specification.new do |s|
   s.name          = "bigdecimal"
-  s.version       = bigdecimal_version
+  s.version       = "3.1.0.dev"
   s.authors       = ["Kenta Murata", "Zachary Scott", "Shigeo Kobayashi"]
   s.email         = ["mrkn@mrkn.jp"]
 
--- a/ext/bigdecimal/extconf.rb
+++ b/ext/bigdecimal/extconf.rb
@@ -3,10 +3,7 @@ require 'mkmf'
 
 def check_bigdecimal_version(gemspec_path)
   message "checking RUBY_BIGDECIMAL_VERSION... "
-
-  bigdecimal_version =
-    IO.readlines(gemspec_path)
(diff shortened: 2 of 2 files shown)


mrkn:
It seems a bug of bundler, isn't it?

casperisfine:
I'm not 100% certain, but I think it's because it needs to rewrite the gemspec for it to work as a git gem.

I don't think it's a big deal to make this change on the bigdecimal side, but if you feel strongly about it I can involve rubygems/bundler maintainers.

mrkn:
Merging this isn't no problem in bigdecimal.

But I guess that it can be an  important topic for bundler users that whether bundler permits using local variables in a gemspec file.

casperisfine:
> But I guess that it can be an important topic for bundler users that whether bundler permits using local variables in a gemspec file.

I know I should probably investigate this myself, but just in case: @deivid-rodriguez is this a known limitation, or something that theoretically could be supported?

deivid-rodriguez:
I don't see how that could be an issue, can you investigate it, please?

deivid-rodriguez:
Oh wait, I understand now. The problem is not the local variable itself, but the code in extconf that assume the variable will be there. Yeat, that sounds like a problem. Bundler evaluates the gemspec and then actually installs a minified version of it. I _think_ this is done because sometimes gemspec make assumptions about stuff external to the gemspec, for example, `require_relative "lib/<foo>/version"`. In an "installed context", that directory structure won't be there, so bundler evaluates to gemspec in the source repository context, and then writes a "serialized version" of it in the installation location.

casperisfine:
Yes, that's what I figured.

byroot:
Based on:

> Merging this isn't no problem in bigdecimal.

And @deivid-rodriguez 's answer, I'll take the liberty to merge.

mrkn:
@byroot You aren't a maintainer of this repository so you must not merge any commit.  That is the regulation.

casperisfine:
@mrkn apologies, I won't do it again.

mrkn:
Could you please ping me when you want to merge some branches.

PR #225, opened 2022-04-13
Fix docs

peterzhu2118:
rdoc parses f[i] as a link, which results in a broken link.


olleolleolle (approved):
This is kind and helpful fixing!

PR #231, opened 2022-06-28
Add support for tangent function

rhannequin:
## Context

`BigMath` module supports sine (`sin`) and cosine (`cos`) trigonometric functions. It also supports arctangent (`acan`). But it doesn't support tangent, which is supported in `Math` module.

## Solution

`tan` is related to `sin` and `cos` function such as ([source](https://en.wikipedia.org/wiki/Trigonometric_functions#Right-angled_triangle_definitions)):
```
tan(x) = sin(x) / cos(x)
```

This PR adds `BigMath.tan(x, precision)`, dependent on existing `BigMath.sin` and `BigMath.cos` methods.


rhannequin:
Hello @mrkn, anything I should do to this PR to improve it?
Please let me know if I missed a step to have this PR reviewed 🙏 
Thank you.

test/bigdecimal/test_bigmath.rb
```diff
@@ -53,6 +53,16 @@ def test_cos
     assert_in_delta(0.0, cos(PI(N) * BigDecimal("301.5"), N))
   end
 
+  def test_tan
+    assert_in_delta(0.0, tan(BigDecimal("0.0"), N))
+    assert_in_delta(0.0, tan(PI(N), N))
+    assert_in_delta(1.0, tan(PI(N) / 4, N))
+    assert_in_delta(Math.sqrt(3.0), tan(PI(N) / 3, N))
```
mrkn, line 60: Use `BigMath.sqrt(BigDecimal(3), N)` instead of `Math.sqrt(3.0)`.
Moreover, I guess we need to use `assert_equal` instead of `assert_in_delta` for all the assertions in this test method.

mrkn: Note that I think we must replace `assert_in_delta` with `assert_equal` in all the other test methods, but they are out of this PR's target.

mrkn (changes requested):
@rhannequin Thank you for your cooperation. I left some comments. Could you please check them?

test/bigdecimal/test_bigmath.rb
mrkn: I want to check the `10*N`-digit case in addition to the `N`-digit case for π/3.  So could you add the following assertion?

```suggestion
    assert_in_delta(Math.sqrt(3.0), tan(PI(N) / 3, N))
    assert_equal(sqrt(BigDecimal(3), 10*N), tan(PI(10*N) / 3, 10*N))
```

rhannequin: Hello @mrkn, thank you for your review :pray: 

I used `assert_in_delta` for your suggestion because the definition of `tan` (`sin / cos`) makes the two methods' precision sum:

``rb
require "bigdecimal/math"

BigMath.sin(BigDecimal("1"), 10)
# => 0.84147098480789650665250231788602089836677454e0

BigMath.cos(BigDecimal("1"), 10)
# => 0.540302305868139717400936607637646275593203e0

BigMath.sin(BigDecimal("1"), 10) / BigMath.cos(BigDecimal("1"), 10)
# => 0.1557407724654902230506974799967261008769125258846034219578498192680719427656331763435628519e1
```

Using `assert_equal` in this case fails as the result has a lot more precision than `N`.

Do you think we should change the `tan` definition to handle precision differently?

Comment (mrkn) [reply test/bigdecimal/test_bigmath.rb]: > Do you think we should change the tan definition to handle precision differently?

We must produce at least `N` precise digits for `BigMath.tan(x, N)`, so the answer is yes.
Instead, I think we can use `assert_in_delta` with specifying `BigDecimal("1e-#{N+1}")` in the `delta` argument.

Comment (mrkn) [reply test/bigdecimal/test_bigmath.rb]: Oops, I'm sorry that I didn't consider how `assert_in_delta` is implemented.
`assert_in_delta` convert the arguments to Float, so we cannot use this to compare BigDecimal values!!

We need to prepare the BigDecimal version of `assert_in_delta` at first.

Comment (mrkn) [reply test/bigdecimal/test_bigmath.rb]: To compare at least `N` digits, the following assertion may be enough:

```
y = BigDecimal("1.73205080756887729352744634150587236694280525381038")
assert_equal(y, BigMath.tan(BigMath.sqrt(3), 50).round(50, :truncate))
```

Comment (rhannequin) [reply test/bigdecimal/test_bigmath.rb]: Hello @mrkn 
Thank you again for reviewing this PR.

I had a look a other tests on the gem and I found that we use `assert_in_delta` in most of them, including `cos` and `sin`.
I understand that this may not be exactly what we're supposed to use for more accurate testing. Although, do you think we should first add this missing method (I'm also planning on adding `acos` and `asin` right after) the way the rest of the gem was implemented?

I'll be happy to help in the same time on updating gem to `assert_in_delta` in a way that is suitable for BigDecimal.
Please let me know what you think.

Comment (mrkn) [reply test/bigdecimal/test_bigmath.rb]: The reason why I'm sticking to a precise alternative of `assert_in_delta` is that your implementation of `tan` depends on `sin` and `cos`, and they are tested by `assert_in_delta`.

If `tan` and other new methods you plan won't depend on other methods tested by `assert_in_delta`, it can be merged.

**Comment by mrkn** [comment]
I'm working on supporting BigDecimal in `assert_in_delta` family assertions at https://github.com/test-unit/test-unit/pull/218.

File: test/bigdecimal/test_bigmath.rb (hunk 1 above)
Comment (rhannequin) [reply test/bigdecimal/test_bigmath.rb]: Thank you for this explanation and the link to the work on `test-unit`, I see better why it is important to wait for a prior work.
I think it's better to stick to a definition of `tan` based on `sin` and `cos` for the simplicity of the gem (of course I'm all ears to counter arguments).

How can I help on your current PR on `test-unit`? This is a bit blurry to me but I guess I have to start somewhere.

Comment (mrkn) [reply test/bigdecimal/test_bigmath.rb]: Fixing `assert_in_delta` is almost done.  After the new version of `test-unit` is released, I will update the master branch of this repository.  And then, please rebase this branch and fix the test.

> I think it's better to stick to a definition of `tan` based on `sin` and `cos` for the simplicity of the gem (of course I'm all ears to counter arguments).

If you can implement `tan` as a series that converges faster than `sin` and `cos`, using the series is better.  If not, you can at least depend on only `sin` by applying basic trigonometric formula `cos(x) = 1 - sin(x)**2`.  Could you please take and compare benchmark results of both `sin/cos` and `sin/(1 - sin**2)` to decide which is better?

Comment (mrkn) [reply test/bigdecimal/test_bigmath.rb]: Additionally, we can treat the case of $x = 0$ as the special case that just returns 0.

**Comment by mrkn** [comment]
I'm working on adding tests with higher precision at https://github.com/ruby/bigdecimal/pull/238

**Comment by rhannequin** [comment]
@mrkn Hello, I'm back after a small break. Maybe I can help with #238?

**Comment by mrkn** [comment]
@rhannequin Although test-unit now supports BigDecimal in `assert_in_delta` family assertions, as tests of BigDecimal need to run also in Ruby's test-all, I have to let the test framework in Ruby's test-all support the new `assert_in_delta` family assertions.

Until finishing this work, could you try to rewrite `tan` implementation to depend on only `sin`?

**Comment by rhannequin** [comment]
Hello @mrkn, I hope you're doing great.

I did some research for a new algorithm for `cos` and arrived to the conclusion to just duplicate it, for now, so that `tan` only depends on `sin`.

I committed it a few minutes ago, what do you think?

File: lib/bigdecimal/math.rb [hunk 2]
```diff
@@ -132,6 +133,54 @@ def cos(x, prec)
...
+      sign = -sign
+      x1  = x2.mult(x1,n)
+      i  += two
+      z  *= (i-one) * i
+      d   = sign * x1.div(z,m)
+      cosine += d
+    end
+
+    sine / cosine
+  end
```
mrkn, line 182: Could you please mention the origin such as the paper reference of this algorithm in the comment?
And, I want to know how this algorithm is better than just doing `sine/(1 - sine**2)`.  Could you explain?

rhannequin:
@mrkn Hello, sorry for taking _so long_ to move forward with this pull request.

My math background is not really advanced and I wasn't familiar with this definition of cosine based on sine.
I changed the implementation of `tan` to only depend on `sin` with this new definition, which is indeed much clearer and probably much faster.

Would you still like a link to a paper or an "official" definition for the documentation?

rhannequin:
@hsbt I allow myself to ping you here as you seem to have been active on the project recently.
Please let me know if this PR needs to be reviewed by someone in particular, or if anything must be changed in the content.

mrkn:
@rhannequin He’s recent work is only for leaving this library out of Ruby’s default gems.  I think there are few people who can review new features of bigdecimal in the Ruby core team.  Could you please wait until I can look into your new commits?

rhannequin:
Absolutely, @mrkn. Sorry if my comment sounded pressing, I wasn't sure you had time for this at the moment. I'll wait now all the time needed.

mrkn:
@rhannequin Thank you for your understanding and cooperation!

lib/bigdecimal/math.rb
nobu: @mrkn I don't believe <tt>tan<var>θ</var></tt> is defined as <tt>sin<var>θ</var> / (1 - sin<sup>2</sup><var>θ</var>)</tt>, but should be <tt>sin<var>θ</var> / cos<var>θ</var></tt> and <tt>cos<var>θ</var></tt> is <tt>√(1 - sin<sup>2</sup><var>θ</var>)</tt>.

mrkn: @nobu The current code was updated against my comment above. Please compare the commit timestamp and the comment timestamp.

mrkn:
Since bigdecimal is now a bundled gem, I think we only need to update the version of test-unit to ensure that the tests can verify the full digits.

rhannequin:
Hello team 👋 
I know I said I would wait all the time needed, but I just wanted an update as a review was requested more than a year ago.
Is there anything preventing this from being merged?
Thank you, and I'll be available for anything that need a change on the PR.

lib/bigdecimal/math.rb
tompng: `sqrt(1 - sine**2, prec)` is not `cosine`. It is `cosine.abs`
Test case of `x` around `-PI/4`, `3*PI/4`, `-3*PI/4` are missing.

test/bigdecimal/test_bigmath.rb
```diff
@@ -53,6 +53,17 @@ def test_cos
...
 
+  def test_tan
+    assert_in_delta(0.0, tan(BigDecimal("0.0"), N))
+    assert_in_delta(0.0, tan(PI(N), N))
+    assert_in_delta(1.0, tan(PI(N) / 4, N))
+    assert_in_delta(sqrt(BigDecimal(3), N), tan(PI(N) / 3, N))
+    assert_in_delta(sqrt(BigDecimal(3), 10 * N), tan(PI(10 * N) / 3, 10 * N))
+    assert_in_delta(0.0, tan(-PI(N), N))
+    assert_in_delta(-1.0, tan(-PI(N) / 4, N))
+    assert_in_delta(-sqrt(BigDecimal(3), N), tan(-PI(N) / 3, N))
```
mrkn, line 64: We need to specify an appropriate BigDecimal for the `delta` parameter in each assertion.
I'll rewrite them later.

lib/bigdecimal/math.rb
```diff
@@ -131,6 +132,26 @@ def cos(x, prec)
...
+  #   BigMath.tan(BigDecimal("0.0"), 4).to_s
+  #   #=> "0.0"
+  #
+  #   BigMath.tan(BigMath.PI(4) / 4, 4).to_s
+  #   #=> "0.999999999999999999999955588155008544487055622030633191403625599381672572e0"
+  #
+  def tan(x, prec)
+    sine = sin(x, prec)
+    cosine = sqrt(1 - sine**2, prec)
+    sine / cosine
```
mrkn, line 152: As I wrote in #319, I want to redefine `sin` by $\sin x = \sqrt{1 - \cos^2 x}$ for $|x| \le \pi$.
So I want to rewrite these lines as follow:

```rb
    x, sign = adjust_x_and_detect_sgin_of_tangent(x)
    t = guarantee_precision(prec) do |n|
      c = cos(x, n)
      s = sqrt(1 - c**2, n)
      s.div(c, n)
    end
    sign >= 0 ? t : -t
```

The `adjust_x_and_detect_sign_of_tangent` method adjusts $x$ within the range $[-\pi, \pi]$ and calculates the result sign from the original $x$.
The `guarantee_precision` method performs Ziv's loop to ensure the result of the block has the desired precision.

mrkn:
Thanks for updating this branch, @rhannequin.  We'll compare this with #321 to see which approach works best.   Whichever solution we choose, I'll make sure you and @tompng are credited as contributors.

PR #232, opened 2022-07-05
Updated license information so it matches the license file

TWinsnes:
This should now be able to be automatically picked up by license checking tools like `licensed`


bigdecimal.gemspec
```diff
@@ -9,7 +9,7 @@ Gem::Specification.new do |s|
   s.summary       = "Arbitrary-precision decimal floating-point number library."
   s.description   = "This library provides arbitrary-precision decimal floating-point number class."
   s.homepage      = "https://github.com/ruby/bigdecimal"
-  s.license       = "Ruby"
+  s.license       = "bsd-2-clause"
```
hsbt, line 12:
```suggestion
  s.license       = ["Ruby", "bsd-2-clause"]
```

Davidslv:
I'd like to check on thing. The [LICENSE.txt](https://github.com/ruby/bigdecimal/blob/master/LICENSE.txt#L45) file states a LEGAL file, is there one? Thank you

bigdecimal.gemspec
TWinsnes: I don't think ruby is a license

Ruby the language uses bsd-2-clause as it's license
https://www.ruby-lang.org/en/about/license.txt

PR #233, opened 2022-07-14
Remove checks for `struct RRational` and `struct RComplex`

nobu:
These are used to see only if `RRATIONAL` and `RCOMPLEX` are available, however, these two are macros and can be checked with `#ifdef` directly.


mrkn (approved):
Looks good.

PR #235, opened 2022-07-29
Document precision=0 and ndigits=0 for converting from Float

mrzasa:
1. The documentation of `Float#to_d` still mentions the old default value of `precision` (`Float::DIG`). I've updated it and described what does the `precision=0` mean.
2. The description of `ndigits=0` disappeared from the documentation of `BigDecimal` function. I've added a short mention there too.

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -3500,6 +3500,9 @@ rb_convert_to_BigDecimal(VALUE val, size_t digs, int raise_exception)
  *  in the value, the result is rounded to that number of digits,
  *  according to the current rounding mode; see BigDecimal.mode.
  *
+ *  When +ndigits+ is 0, the number of digits to correctly represent a float number
+ *  is determined automatically.
+ *
  *  Returns +value+ converted to a \BigDecimal, depending on the type of +value+:
  *
  *  - Integer, Float, Rational, Complex, or BigDecimal: converted directly:
--- a/lib/bigdecimal/util.rb
+++ b/lib/bigdecimal/util.rb
@@ -33,12 +33,16 @@ class Float < Numeric
   #
   # Returns the value of +float+ as a BigDecimal.
   # The +precision+ parameter is used to determine the number of
-  # significant digits for the result (the default is Float::DIG).
+  # significant digits for the result. When +precision+ is set to +0+,
+  # the number of digits to represent the float being converted is determined
+  # automatically.
+  # The default +precision+ is +0+.
   #
   #     require 'bigdecimal'
   #     require 'bigdecimal/util'
   #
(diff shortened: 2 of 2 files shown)


ext/bigdecimal/bigdecimal.c
```diff
@@ -3500,6 +3500,9 @@ rb_convert_to_BigDecimal(VALUE val, size_t digs, int raise_exception)
  *  in the value, the result is rounded to that number of digits,
  *  according to the current rounding mode; see BigDecimal.mode.
  *
+ *  When +ndigits+ is 0, number of digits to correctly represent a float number
```
olleolleolle, line 3503:
```suggestion
 *  When +ndigits+ is 0, the number of digits to correctly represent a float number
```

Adding this "the" makes this harmonize with the other file's change in wording.

mrzasa: Thanks!

ext/bigdecimal/bigdecimal.c
```diff
@@ -3500,6 +3500,9 @@ rb_convert_to_BigDecimal(VALUE val, size_t digs, int raise_exception)
  *  in the value, the result is rounded to that number of digits,
  *  according to the current rounding mode; see BigDecimal.mode.
  *
+ *  When +ndigits+ is 0, the number of digits to correctly represent a float number
+ *  is determined by a +dtoa+ algorithm (see more in C function +dtoa+ with +mode=0+).
```
mrkn, line 3504: Could you please remove the mention of `dtoa`.  This is not the specification but just the current implementation detail.

lib/bigdecimal/util.rb
```diff
@@ -33,12 +33,16 @@ class Float < Numeric
   #
   # Returns the value of +float+ as a BigDecimal.
   # The +precision+ parameter is used to determine the number of
-  # significant digits for the result (the default is Float::DIG).
+  # significant digits for the result. When +precision+ is set to +0+,
+  # the number of digits to represent the float being converted is determined
+  # automatically (see more in C function +dtoa+ with +mode=0+).
```
mrkn, line 38: The same as above.

mrkn (changes requested):
@mrzasa I've left comments.  Please check them.

ext/bigdecimal/bigdecimal.c
mrzasa: done

lib/bigdecimal/util.rb
mrzasa: done

mrzasa:
I've removed mentions of `dtoa` from the docs.

mrkn:
@mrzasa Thank you very much.

PR #237, opened 2022-09-19
Remove definitions in missing.h for old Rubies

peterzhu2118:
These definitions are not needed as all recent (>= 2.3) Rubies have them defined.


mrkn:
Thank you very much!

PR #238, opened 2022-10-08
Add precise tests

mrkn:


tompng:
Precision test added in #414

PR #240, opened 2022-11-13
Fix divison

### LINKED ISSUE #222: Exponential growth in precision causing memory usage to spike as of 3.1.2
I upgraded to BigDecimal 3.1.2 to pull in the fix to #220, which was causing crashes in our Rails app. However, I believe the fix applied in 127a1b5 created the potential for exponential growth in the number of digits a BigDecimal can have. The result is that certain calculations can eat up all available memory (several gigs) and crash the process.

Take the following as an example, which outputs the number of digits from the calculation:

```ruby
Array.new(10) { BigDecimal("1.03") }.inject(1.0) { |result, num| result / num }.to_digits.length
```

In 3.1.1, this calculation produces 119 digits.

In 3.1.2, this calculation produces 23,033 digits.

Moreover, if `Array.new(10)` is grown to `Array.new(20)`, 3.1.1 produces 209 digits, whereas 3.1.2 produces 23,592,953.

The runaway growth in digits and resulting memory spike could be solved by rounding the result in inject's block like so:

```ruby
Array.new(10) { BigDecimal("1.03") }.inject(1.0) { |result, num| (result / num).round(3) }.to_digits.length
```

But others may run into this same surprise. We've also run into `PG::NumericValueOutOfRange` exceptions from postgres since upgrading to 3.1.2, and my hunch is the same root issue is at play.

Let me know if I can provide any more details.

mrkn:
This will fix #222.

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -1855,22 +1855,37 @@ static VALUE BigDecimal_DoDivmod(VALUE self, VALUE r, Real **div, Real **mod);
  */
 static VALUE
 BigDecimal_div(VALUE self, VALUE r)
-/* For c = self/r: with round operation */
 {
-    ENTER(5);
-    Real *c=NULL, *res=NULL, *div = NULL;
-    r = BigDecimal_divide(self, r, &c, &res, &div);
-    if (!NIL_P(r)) return r; /* coerced by other */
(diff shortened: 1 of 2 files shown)


jeffktan:
Following up on this PR as it causes slowness in our codebase. It may cause slowness many!

mrkn:
#329 fixes the problem.

PR #243, opened 2022-11-30
Handle correctly #reminder with infinity. Fixes #187

### LINKED ISSUE #187: Adjust modulus/remainder calculation for infinity arguments
```ruby
require 'bigdecimal'
bd = BigDecimal("4.2")
bd.remainder(BigDecimal("+Infinity")) # => NaN, should be bd
bd.remainder(BigDecimal("-Infinity")) # => NaN, should be bd
```

See https://bugs.ruby-lang.org/issues/6120

I've submitted a pull request to ruby to fix the issue with Integer/Float, but bigdecimal should be handled separately.

mrzasa:
Fixes https://github.com/ruby/bigdecimal/issues/187 by adding a guard to `BigDecimal_divremain`. This returns early if the divider is infinity.

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -2082,6 +2082,13 @@ BigDecimal_divremain(VALUE self, VALUE r, Real **dv, Real **rv)
     if (!b) return DoSomeOne(self, r, rb_intern("remainder"));
     SAVE(b);
 
+    if (VpIsPosInf(b) || VpIsNegInf(b)) {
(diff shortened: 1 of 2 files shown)


mrzasa:
@mrkn could you take a look? I believe it's ready for review now

mrkn:
@mrzasa Thank you!

nobu:
Doesn't this need to bump the version?
I think [this guard] should use `BigDecimal::VERSION` instead of `RUBY_VERSION`.

[this guard]: https://github.com/ruby/ruby/commit/0d8ef62fc293dc04110f36382a7e8bddec6aee15

PR #248, opened 2023-02-01
Make BigDecimal WB protected

peterzhu2118:
BigDecimal has no references, so it is WB protected.


byroot (approved):
For context we were just looking at the most common unportected objects in our heap dumps, and `BigDecimal` came up second after `Time` with 75k instances (we do lots of money computations).

mrkn:
Thanks!

Issue #169, opened 2020-08-27
Allow installation on JRuby


headius:
JRuby has its own bigdecimal library based on the JDK's built-in BigDecimal. We also do not support C extensions, which means this gem can't be installed at all on JRuby.

This leads to issues when users add bigdecimal gem to their applications and try to migrate to JRuby. This will become more of a problem if downstream libraries also start adding bigdecimal, since there's no way to swap in our implementation.

There are two possible ways forward:

* Modify the gem so it does not attempt to build an extension on JRuby and just falls back on JRuby's built-in bigdecimal support.
* Release a separate gem for JRuby that contains our BigDecimal extension.

Of these, the first is clearly the least amount of work. The second would require us to remove all references from JRuby's core classes to the bigdecimal library, which might break some of our public APIs surrounding converting Java BigDecimal to Ruby and back again.

This is the cause of https://github.com/noahgibbs/rsb/issues/12 and that issue would be resolved if this one is resolved.

jodosha:
👍

hsbt:
@headius Can we add the empty extension for Java platform same with `date` gem? And can we add JRuby specific logic to https://github.com/ruby/bigdecimal/blob/master/lib/bigdecimal.rb?

headius:
@hsbt We would still like to get our implementations of these gems moved into the same repository and released together, but yes for now it would be best if we had some no-op versions that would skip installing the C extension.

At least `bigdecimal` and `openssl` need this treatment now. The -java version of `openssl` should skip installing the extension and depend on our `jruby-openssl` gem.

hsbt:
> The -java version of `openssl` should skip installing the extension and depend on our `jruby-openssl` gem.

This proposal seems good first step. Can you file this to ruby/openssl repo?

headius:
Is it posible to start releasing the -java stub gem, please? This has now caused sequel to break on JRuby, as it depends on the bigdecimal gem and the CRuby extension is not supported.

https://github.com/jeremyevans/sequel/issues/2076

headius:
bigdecimal-java 3.1.4 has been released, so I believe this is resolved! We just need the -java version released for each version of bigdecimal going forward.

headius:
I understand that @mrkn does not want to maintain the JRuby version in this repository, but we still want to spin our code off as a gem.

[...]

PR #250, opened 2023-02-09
Stub out extension build on JRuby

headius:
JRuby currently ships its own internal bigdecimal extension as part of the core libraries. In order for users to be able to add bigdecimal to their Gemfile or gem dependencies, we need to stub out the C extension and just load the extension shipped with JRuby.

In the future we will try to move our BigDecimal implementation into the gem, but for now this is the simplest way to make it installable on JRuby.

See #169

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 gemspec
 
 gem "benchmark_driver"
-gem "fiddle"
+gem "fiddle", platform: :ruby
 gem "rake", ">= 12.3.3"
 gem "rake-compiler", ">= 0.9"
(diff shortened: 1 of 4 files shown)


headius:
Note this should also let us start to source the .rb files from the gem rather than versioning them in the JRuby repository. The last step will be moving our native BigDecimal implementation into the gem, which will hopefully happen some time this year.

mrkn:
> The last step will be moving our native BigDecimal implementation into the gem, which will hopefully happen some time this year.

Migration of the Java native BigDecimal into this repo will force us to handle both C and Java implementations at the same time when we add the new feature and fix the behavior.  I want to avoid this situation.

PR #257, opened 2023-02-21
Bump up to 3.1.4

nobu:
ruby/bigdecimal#187 has changed a behavior and ruby/spec also needed a follow up at ruby/ruby@0d8ef62fc293dc04110f36382a7e8bddec6aee15.
However, because bigdecimal is a separate gem and can be updated in older versions of ruby, `RUBY_VERSION` is not appropriate for this guard.
That means it needs bumped up `BigDecimal::VERSION`.

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -31,7 +31,7 @@
 #include "bits.h"
 #include "static_assert.h"
 
-#define BIGDECIMAL_VERSION "3.1.3"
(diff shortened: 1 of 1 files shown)


casperisfine:
The latest version on rubygems is still 3.1.3 https://rubygems.org/gems/bigdecimal/

Were you planning to publish that release?

mlarraz:
Would it be possible to update the changelog to help projects decide whether or not to upgrade to this version right now?

PR #259, opened 2023-03-17
fix: typo in document comments of `f_BigDecimal` function

honood:
The [documents for `Kernel#BigDecimal` method](https://ruby-doc.org/3.2.1/exts/bigdecimal/Kernel.html#method-i-BigDecimal) has a typo in the second line from the bottom:

> Returns `nil` if keyword argument `exception` is `true`.

which should be:

> Returns `nil` if keyword argument `exception` is `false`.


mrkn:
@honood Good catch! Thank you!!

PR #264, opened 2023-07-01
Add .to_s('F') digit grouping for integer part

cryptogopher:
Right now to_s('F') digit grouping works properly only for fractional part. For integer part result is unspecified. In reality we also get grouping, but of unpractical nature:

```
# ruby -e "require 'bigdecimal'; puts BigDecimal('1234567.1234567').to_s('3F')"
123 456 7.123 456 7
```

This patch fixes grouping of integer part:

```
# ruby -e "require 'bigdecimal'; puts BigDecimal('1234567.1234567').to_s('3F')"
1 234 567.123 456 7
```

I consider this change backwards compatible as:
a) docs say only about grouping fractional part, no declarations about grouping integer part,
b) total string length without changes,
c) integer part grouping didn't make sense so there should be no or not much real life dependencies.

Also added some tests.


test/bigdecimal/test_bigdecimal.rb
```diff
@@ -1415,9 +1415,13 @@ def test_ceil
...
-    assert_equal('-123.45678 90123 45678 9', BigDecimal('-123.45678901234567890').to_s('5F'))
-    assert_equal('+123.45678901 23456789', BigDecimal('123.45678901234567890').to_s('+8F'))
-    assert_equal(' 123.4567890123456789', BigDecimal('123.45678901234567890').to_s(' F'))
+    assert_equal('0.0', BigDecimal('0').to_s)
+    assert_equal('-123 45678 90123.45678 90123 45678 9', BigDecimal('-1234567890123.45678901234567890').to_s('5F'))
+    assert_equal('+12345 67890123.45678901 23456789', BigDecimal('1234567890123.45678901234567890').to_s('+8F'))
+    assert_equal(' 1234567890123.4567890123456789', BigDecimal('1234567890123.45678901234567890').to_s(' F'))
+    assert_equal('100 000 000 000.000 000 000 01', BigDecimal('100000000000.00000000001').to_s('3F'))
+    assert_equal('0.0 0 0 0 0 0 0 0 0 0 0 0 1', BigDecimal('0.0000000000001').to_s('1F'))
+    assert_equal('+1000000 0000000.0', BigDecimal('10000000000000').to_s('+7F'))
```
mrkn, line 1424: @cryptogopher Could you please add the cases where the given argument is invalid?

mrkn (changes requested):

test/bigdecimal/test_bigdecimal.rb
cryptogopher: You mean the argument to `to_s`? And invalid like completely invalid format string? I'm not sure if I understand your request.
As a side note: I didn't change the argument parsing part, only the number to string conversion function `VpToFString`.

mrkn: Oops! I misread the following part of the diff...

image

zzak:
Looks like this broke some Active Support tests:
https://github.com/rails/rails/pull/48693

nobu:
This change does not related to ruby versions.
Before bad habit that testing this by `RUBY_VERSION` will spread more, we should bump up `BigDecimal::VERSION`.

PR #269, opened 2023-09-11
Use test-unit-ruby-core gem

hsbt:
We should use https://github.com/ruby/test-unit-ruby-core instead of vendor files.


hsbt:
Fixed by https://github.com/ruby/ruby/pull/8419

PR #273, opened 2023-11-01
fixed docs for .scale

krtschmr:
the scale of `1` is actually 0


hsbt:
Thanks. 

`BigDecimal("1").scale` returns `0` from Ruby 3.1.

PR #275, opened 2023-11-16
[DOC] Add section Methods for Working with JSON

BurdetteLamar:
The linked methods are weakly documented at present, but stronger doc is already in a PR over at JSON.


peterzhu2118 (approved):

mrkn:
@peterzhu2118 Wait, why did you merge this without the maintainer confirmation?

These are methods added by json.gem, right? So, these description should put in the documentation of json.gem.

mrkn:
Alright, I understand that we need to include them in this repository. Therefore, I think it's important to note that 'json/add/bigdecimal' is provided by the json.gem. What are your thoughts?

peterzhu2118:
I opened #278 to add this.

PR #277, opened 2023-12-06
Add missing documents

nobu:


ext/bigdecimal/bigdecimal.c
```diff
@@ -4238,6 +4256,17 @@ BigDecimal_negative_zero(void)
...
+static inline VALUE
+BigDecimal_literal(const char *str)
+{
+    VALUE arg = rb_str_new_cstr(str);
+    VALUE val = f_BigDecimal(1, &arg, rb_cBigDecimal);
+    rb_gc_register_mark_object(val);
+    return val;
+}
+
+#define BIGDECIMAL_LITERAL(var, val) (BIGDECIMAL_ ## var = BigDecimal_literal(#val))
```
hsbt, line 4268: This and related changes are intentional? It seems accidentally commits.

nobu: It is needed for ["INFINITY"](https://github.com/ruby/bigdecimal/pull/277/files#diff-2bad21ea62ae18c67587f0544f35719cd07a33c83d518f4f91c15fc1f1c70425R4556) and so on, because RDoc requires docs for constants to be placed just before each constant.

PR #280, opened 2023-12-14
Correctly computing loop iterations in `BigDecimal#sqrt`

z2-2z:


mrkn:
@z2-2z Thank you very much!

hsbt:
This change break i686 environment https://github.com/ruby/ruby/actions/runs/7321152433/job/19941861006

mrkn:
@hsbt Did you compile the ruby you used with `-msse2 -mfpmath=sse`?

mrkn:
I'm interested in what is the cause of that failure, but I have no desire to continue full support for the i686 environment.

PR #283, opened 2024-01-17
Workaround for test-bundled-gems at ruby/ruby repo

hsbt:
There is no way to load `bigdecimal` extension in `assert_in_out_err` with `make test-bundled-gems` of ruby/ruby repository.

I added extension directory for that explicitly. I know this is ad-hoc solution. We should provide these paths to like `assert_in_out_err` assertions in `ruby/ruby` side.


hsbt:
https://github.com/ruby/bigdecimal/actions/runs/7552785831/job/20562319734?pr=283 is caused by `io-console` side at https://github.com/ruby/io-console/pull/65.

I will release fixed version tomorrow.

PR #302, opened 2024-09-05
Accept no digits in the fractional part

### LINKED ISSUE #301: Cannot parse '0.E-9'
When working with Apache Arrow Decimals, I found that they use this string representation for zero (with precision 38, scale 9):
```
0.E-9
```
(I have also filed this [issue](https://github.com/apache/arrow/issues/43877) on Apache Arrow repo that includes more context)

But BigDecimal does not recognize `0.E-9`
image

I found that postgres also accepts `0.E-9`
image

Will BigDecimal be able to support such string representation?

kou:
fix #301

For example, "0.E-9" and "0." are accepted.

FYI:

`Float()`/`.to_f` will accept "0.E-9":
https://bugs.ruby-lang.org/issues/20705

The current `Float` doesn't accept "0." but `.to_f` accepts "0.".

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -5277,7 +5277,7 @@ VP_EXPORT Real *
 VpAlloc(size_t mx, const char *szVal, int strict_p, int exc)
 {
     const char *orig_szVal = szVal;
-    size_t i, j, ni, ipf, nf, ipe, ne, dot_seen, exp_seen, nalloc;
+    size_t i, j, ni, ipf, nf, ipe, ne, exp_seen, nalloc;
     size_t len;
     char v, *psz;
     int  sign=1;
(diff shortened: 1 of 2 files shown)


kou:
@mrkn What do you think about this?

mrkn:
@kou I'm in favor of this change, but I want to merge this after Ruby's [[Feature #20705]](https://bugs.ruby-lang.org/issues/20705) is merged.

kou:
@mrkn The ticket was closed.

mrkn:
@kou Thank you very much!

kou:
@mrkn Do you want to release a new version with this so that Ruby 3.4.0 and `BigDecimal()` use the same behavior.

mrkn:
Thank you for reminding me. I will do.

Kenta Murata, sent from Gmail Mobile

mrkn:
@kou The new version was shipped.

kou:
Thanks!

Issue #262, opened 2023-05-17
I would expect BigDecimal#divmod to return type [Int, BigDecimal] (like Integer#divmod, Float#divmod)


michaelglass:
`Integer#divmod` returns type `[Integer, Integer]`
`Float#divmod` returns type `[Integer, Float]`
`BigDecimal#divmod` returns type `[BigDecimal, BigDecimal]`

Which feels surprising.

```
1.divmod(2)
=> [0, 1]
(1.0).divmod(2)
=> [0, 1.0]
BigDecimal(1).divmod(2)
=> [0.0, 0.1e1]
```

(thank you BigDecimal maintainers, you are wonderful)

PR #312, opened 2025-02-04
Cast divmod quotient to int

mrzasa:
Fixes https://github.com/ruby/bigdecimal/issues/262
(description in TBD)


mrkn:
@mrzasa We should consider the timing of merging this change into the master branch due to the incompatibility introduced. It should be v3.4 or v3.5. Wait a moment, please.

Issue #213, opened 2021-12-03
BigDecimal should accept a Float value without a precision


mrkn:
```
irb(main):002:0> BigDecimal(5.1)
(irb):2:in `BigDecimal': can't omit precision for a Float. (ArgumentError)
        from (irb):2:in `<main>'
        from /home/mrkn/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/irb-1.3.8.pre.9/exe/irb:11:in `<top (required)>'
        from /home/mrkn/.rbenv/versions/3.0/bin/irb:23:in `load'
        from /home/mrkn/.rbenv/versions/3.0/bin/irb:23:in `<main>'
```

This should works like below.

```
irb(main):003:0> BigDecimal(5.1, Float::DIG+1)
=> 0.51e1
```

~This is necessary to resolve #212~.

vanboom:
:+1:   Why require precision?   Can't it just work like it did in previous ruby versions?

PR #314, opened 2025-02-04
Allow BigDecimal accept Float without precision

mrzasa:
Make sure that conversion from a Float works consistently with the conversion from a string representing that float.

Current `master` state:
```
pry(main)> BigDecimal(1.2)
ArgumentError: can't omit precision for a Float.
from (pry):22:in `BigDecimal'
pry(main)> BigDecimal('1.2')
=> 0.12e1
```

After this changes:
```
pry(main)> BigDecimal(1.2)
=> 0.12e1
pry(main)> BigDecimal('1.2')
=> 0.12e1
```

When precision is omitted it's set to `0`.

Fixes https://github.com/ruby/bigdecimal/issues/213
Blocked by https://github.com/ruby/bigdecimal/pull/313

### DIFF
--- a/test/bigdecimal/test_bigdecimal.rb
+++ b/test/bigdecimal/test_bigdecimal.rb
@@ -166,7 +166,8 @@ class TestBigDecimal < Test::Unit::TestCase
     assert_equal(BigDecimal("0.1235"), BigDecimal(0.1234567, 4))
     assert_equal(BigDecimal("-0.1235"), BigDecimal(-0.1234567, 4))
     assert_equal(BigDecimal("0.01"), BigDecimal(0.01, Float::DIG + 1))
-    assert_raise_with_message(ArgumentError, "can't omit precision for a Float.") { BigDecimal(4.2) }
+    assert_nothing_raised { BigDecimal(4.2) }
+    assert_equal(BigDecimal(4.2), BigDecimal('4.2'))
     assert_raise(ArgumentError) { BigDecimal(0.1, Float::DIG + 2) }
     assert_nothing_raised { BigDecimal(0.1, Float::DIG + 1) }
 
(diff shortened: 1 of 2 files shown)


test/bigdecimal/test_bigdecimal.rb
```diff
@@ -166,7 +166,8 @@ def test_BigDecimal_with_float
     assert_equal(BigDecimal("0.1235"), BigDecimal(0.1234567, 4))
     assert_equal(BigDecimal("-0.1235"), BigDecimal(-0.1234567, 4))
     assert_equal(BigDecimal("0.01"), BigDecimal(0.01, Float::DIG + 1))
-    assert_raise_with_message(ArgumentError, "can't omit precision for a Float.") { BigDecimal(4.2) }
+    assert_nothing_raised { BigDecimal(4.2) }
+    assert_equal(BigDecimal(4.2), BigDecimal('4.2'))
```
mrkn, line 170: @mrzasa Could you add more assertions to check many cases, especially for numbers that don't have accurate representations in Float?
Moreover, we must compare two values by `assert_in_delta` and pass an appropriate value to the `delta` argument for the original Float value; it should be 0.5ulp.

mrkn: I reconsidered how to check here.

I suggest comparing the result of `to_f` with the original Float value, like this: `assert_equal(4.2, BigDecimal(4.2).to_f)`.
This approach is preferable because it doesn’t depend on the specific decimal representation for a given Float value. I’d like to avoid relying on the particular algorithm used for converting a Float to a BigDecimal.

tompng: Adding a same test case as `BigDecimal(float, 0)` is enough.
But since test for `BigDecimal(float, 0)` is currently missing, adding both will be nice.

```ruby
assert_equal(BigDecimal("string representation"), BigDecimal(float_literal_here, 0))
assert_equal(BigDecimal("string representation"), BigDecimal(float_literal_here))

# Just like this one on line 166-168
assert_equal(BigDecimal("0.1235"), BigDecimal(0.1234567, 4))
```

tompng (approved):
Thank you. looks good :+1:

Perhaps some tests for `BigDecimal(x, various_number)` are missing, some test are too strict.
```ruby
# this test is too strict
assert_equal(BigDecimal("0.01"), BigDecimal(0.01, Float::DIG + 1))
# because:
BigDecimal(0.07, Float::DIG + 1) #=> 0.7000000000000001e-1
```
but let's consider it a separate issue of test.

PR #316, opened 2025-05-05
Add a precision assertion to BigMath test

tompng:
Precision assertion in BigMath test is missing.
BigMath uses `BigDecimal.double_fig` (16 in my env), so the calculated precision is normally >=16.
We need to test with precision several times larger than 16 to ensure that the result is really precise enough.

This assertion will be helpful when adding new methods to BigMath (example: `BigMath.tan`)

## Precision check
Assume `BigMath.func(v, 2 * prec)` to be the expected result which is precise enough and calculates the difference.
`assert_fixed_point_precision` is for sin and cos. `assert_relative_precision` is for other methods.

## sin and cos

I think `BigMath.sin(x)`, `BigMath.cos(x)` with large `x` is not precise enough. This test case is not added in this pull request.
```ruby
BigMath.sin(BigDecimal('1e50'), 100) - BigMath.sin(BigDecimal('1e50'), 200)
# => -0.7388355386-67 # precision should be >=100 but is 67
```

### DIFF
--- a/test/bigdecimal/test_bigmath.rb
+++ b/test/bigdecimal/test_bigmath.rb
@@ -6,29 +6,12 @@ class TestBigMath < Test::Unit::TestCase
   include TestBigDecimalBase
   include BigMath
   N = 20
-  N_LARGE = 100
   # SQRT in 116 (= 100 + double_fig) digits
   SQRT2 = BigDecimal("1.4142135623730950488016887242096980785696718753769480731766797379907324784621070388503875343276415727350138462309123")
   SQRT3 = BigDecimal("1.7320508075688772935274463415058723669428052538103806280558069794519330169088000370811461867572485756756261414154067")
   PINF = BigDecimal("+Infinity")
   MINF = BigDecimal("-Infinity")
   NAN = BigDecimal("NaN")
-
-  def assert_fixed_point_precision(n = N_LARGE)
-    value = yield(n)
-    expected = yield(2 * n)
-    precision = -(value - expected).exponent
-    assert(value != expected, "Unable to estimate precision for exact value")
-    assert(precision >= n, "Precision is not enough: #{precision} < #{n}")
-  end
-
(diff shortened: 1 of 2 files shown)


mrkn:
👍

test/bigdecimal/test_bigmath.rb
```diff
@@ -6,10 +6,30 @@ class TestBigMath < Test::Unit::TestCase
   include TestBigDecimalBase
   include BigMath
   N = 20
+  N_LARGE = 100
```
mrkn, line 9: This is just a comment to share my concern regarding the use of BigDecimal on 64-bit architectures. Each BigDecimal digit element can hold 19 decimal digits, meaning that 100 would require 6 BigDecimal digits. We should consider if 6 BigDecimal digits is truly a large number.

test/bigdecimal/test_bigmath.rb
```diff
@@ -6,10 +6,30 @@ class TestBigMath < Test::Unit::TestCase
...
+    expected = yield(2 * n)
+    precision = -(value - expected).exponent
+    assert(value != expected, "Unable to estimate precision for exact value")
+    assert(precision >= n, "Precision is not enough: #{precision} < #{n}")
+  end
+
+  def assert_relative_precision(n = N_LARGE)
+    value = yield(n)
+    expected = yield(2 * n)
+    precision = -(value / expected - 1).exponent
```
mrkn, line 28: Don’t you have to specify the quotient precision explicitly in the division?

mrkn (changes requested):

test/bigdecimal/test_bigmath.rb
tompng: Update the precision assertion to use [50, 100, 150]
Trying multiple precisions will increase the probability of detecting mis-calculation.

PR #317, opened 2025-05-11
Ensure BigMath.sin and BigMath.cos to be within -1..1

tompng:
BigMath.sin and cos was returning `> 1` or `<-1` value in few cases.
```ruby
irb(main):001> BigMath.sin(BigMath::PI(30) / 2, 30)
=> 0.1000000000000000000000000000000000000000000000099061665879329355e1
irb(main):002> BigMath.cos(BigMath::PI(30), 30)
=> -0.1000000000000000000000000000000000000000000004381047173448900999e1
```

The precision is correct (`error < 1e-44`), but sin and cos is often assumed to be within `-1..1`.
```ruby
sin = BigMath.sin(x)
cos = (1 - sin**2) ** 0.5 # assume sin**2 to be <= 1, but it was not ensured
```


mrkn:
@tompng, it may be advisable to apply a provisional adjustment so the result falls within each function’s valid range.
We should be mindful of loss of precision when $|x|$ approaches to $\pi k/2$ for `sin` and $k\pi$ for `cos`, respectively.
Could you investigate the extent of the loss of precision in these regions?

tompng:
There is no loss of accuracy. Error decreases in every case.
```ruby
if master_branch_sin(x).abs <= 1
  # no change, no accuracy loss
  return master_branch_sin(x)
elsif master_branch_sin(x) > 1
  return 1
  # true_sin(x) <= this_pull_req_sin(x) == 1 <= master_branch_sin(x), so the error always decrease
elsif master_branch_sin(x) < -1
  return -1
  # master_branch_sin(x) < this_pull_req_sin(x) == 1 <= true_sin(x)
end
```

Issue #81, opened 2017-11-30
BigMath.tan


mrkn:
Original issue is https://bugs.ruby-lang.org/issues/3269.

im-prince:
any one working on this issue if not please assign me

PR #321, opened 2025-05-17
Implement BigMath.tan

tompng:
#81

Calculate `tan(x, prec)` by `sin(x, sin_prec) / cos(x, cos_prec)`.
If the precision of sin/cos is not enough, increase prec and recalculate again.

### Example of increasing precision of cosine
```ruby
x = BigMath::PI(150) / 2
BigMath.tan(x, 100)

# First, calculate cos with denominator_prec = 116
cos(x, 116) #=> -0.57442751e-131
# This value is not correct at all. We can only know that |cos(x, 116)| < 1e-116.
# No information to estimate appropriate denominator_prec, so multiply denominator_prec by 3/2 and recalculate

116 * 3 / 2 #=> 174
cos(x, 174) #=> -0.133945260097112077688525522e-165

# This value is inaccurate. Only the first 174-165 = 9 digits can be trusted.
# We need at least 91 more digits, so increase denominator_prec by 91+double_fig and recalculate

174 + 91 + 16 #=> 281
cos(x, 281) #=> -0.133945260097112077688525577-165

# Now precision of cos is enough to calculate sin(x)/cos(x)
return sin(x, 100) / cos(x, 281) #=> -0.7465736370775545e166
```

### About sin and cos precision
```ruby
BigMath.sin(around_zero, prec)
#=> relative precision. `(calculated_value/true_value - 1).abs < 10**(-prec)` seems to be ensured

BigMath.sin(other_value, prec)
#=> fixed point precision. `(calculated_value - true_value).abs < 10**(-prec)` is ensured

BigMath.cos(other_value, prec)
#=> fixed point precision. `(calculated_value - true_value).abs < 10**(-prec)` is ensured
```

### DIFF
--- a/lib/bigdecimal/math.rb
+++ b/lib/bigdecimal/math.rb
@@ -7,6 +7,7 @@ require 'bigdecimal'
 #   sqrt(x, prec)
 #   sin (x, prec)
 #   cos (x, prec)
+#   tan (x, prec)
 #   atan(x, prec)
 #   PI  (prec)
 #   E   (prec) == exp(1.0,prec)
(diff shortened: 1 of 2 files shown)


tompng:
I think `sin` and `cos` should also calculate in relative precision.
`BigMath.tan(x,prec)` is (and should be) relative precision in all range. If sin and cos is still fixed-point precision, it's not consistent, misleading, can cause bug.

After #321, `def tan(x, prec) = sin(x, prec)/cos(x, prec)` is enough. The way this pull request implements tan is too complicated.

PR #322, opened 2025-05-17
Simplify power by integer logic

tompng:
Refactor and reduce the number of multiplication.

Example calculation of `x**16`
```ruby
x * x**8 * x**4 * x**2 * x # x**8, x**4, x**2 are calculate separately
# ↓
x * x * x**2 * x**4 * x**8 # result of x**4 is re-used to calculate x**8
```

Calculation of the largest `x**n` is dominant. Although multiplication is reduced, the performance improvement is limited.

### DIFF
--- a/test/bigdecimal/test_bigdecimal.rb
+++ b/test/bigdecimal/test_bigdecimal.rb
@@ -1660,8 +1660,7 @@ class TestBigDecimal < Test::Unit::TestCase
     BigDecimal.save_limit do
       BigDecimal.limit(1)
       x = BigDecimal("3")
-      assert_equal(90, x ** 4) # OK? must it be 80?
-      # 3 * 3 * 3 * 3 = 10 * 3 * 3 = 30 * 3 = 90 ???
+      assert_equal(80, x ** 4)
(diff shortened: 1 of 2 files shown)


test/bigdecimal/test_bigdecimal.rb
```diff
@@ -1660,8 +1660,7 @@ def test_limit
     BigDecimal.save_limit do
       BigDecimal.limit(1)
       x = BigDecimal("3")
-      assert_equal(90, x ** 4) # OK? must it be 80?
-      # 3 * 3 * 3 * 3 = 10 * 3 * 3 = 30 * 3 = 90 ???
```
tompng, line 1664: Before:  `3**4` → `(3 * (3**2)) * 3`  → `(3 * 9) * 3` → `round(27) * 3` → `90`
After: `3**4` → `(3 * 3) * (3**2)` → `9 * 9` → `round(81)` → `80`

Although this test case seems to be improved, just a coincidence.
This pull request is not a bug fix. Large error still exist.
```ruby
BigDecimal.limit 1
(BigDecimal('2')**100).to_i / 2**100 #=> 1577
```

Issue #319, opened 2025-05-15
Development plan


mrkn:
I opened this issue to outline the development plan for bigdecimal, especially for our new co-maintainer, @tompng. Thank you, @tompng, for joining the project.

Over the past several years, I have been largely inactive in bigdecimal development because I rarely have more than 30 minutes of free time each day for open-source work as hobby. Two years ago, I expected to resume development within 2-3 years, but the current constraints are likely persist for at least another three years. Consequently, I have decided to share my thoughts on bigdecimal's roadmap here.

The roadmap below is compiled from my notes and memory. Because there are many tasks I wanted to do and I want to record them accurately here, I will continue refining this issue over the coming days.

In the first iteration, I describe only the highest-priority items; the others are merely listed for reference. I will update this issue several times as I flesh out the remaining topics.

## Basic policy

Development must follow these principles:

- Avoid converting values to base-2 representations for computation whenever possible.
- Minimize CPU time and memory usage.
- Keep external dependencies to an absolute minimum.

### Tests for high-precision values
Verify each high-precision result with either [`assert_in_delta`](https://test-unit.github.io/test-unit/en/Test/Unit/Assertions.html#assert_in_delta-instance_method)  or [`assert_in_epsilon`](https://test-unit.github.io/test-unit/en/Test/Unit/Assertions.html#assert_in_epsilon-instance_method), explicitly passing an appropriate **BigDecimal** to the `delta` or `epsilon` parameter to avoid converting to `Float` and maintain accuracy.

### ✅ ~Improve division performance~
~Resolve issue #222 (details on the current slowdown will be added later).~

### Improve multiplication performance
- Implement Karatsuba multiplication.
- Implement Toom-3 multiplication.
- Implement NTT-based multiplication with 2 or 3 primes.

### Further improve division performance
- Implement Karatsuba division.

### Improve sqrt performance
- Implement Karatsuba sqrt.

### Trigonometric functions 
- Increase the accuracy of `cos` and eliminate unnecessary extra-digit computation.
- Use the identity $\cos(x) = \cos(x - \pi)$ for $|x| \ge \pi$.
- Rewrite `sin` as $\sin(x) = \sqrt{1 - \cos^2(x)}$ except when $x$ is near zero.
- Adopt Ziv's loop in the proposed `tan` implementation (see #231) to guarantee precision.

### Support all the functions in Math module
(I'll outline this later)

### etc.
(I'll add additional TODO items and describe them in detail later)

[...]

PR #323, opened 2025-05-17
Rewrite BigDecimal#sqrt in ruby using Integer.sqrt

tompng:
This approach violates the policy in #319
> Avoid converting values to base-2 representations for computation whenever possible.

But I think that both performance gain and maintainability gain might be worth enough.

## Using `Integer.sqrt`

Ruby's `Integer.sqrt` is pretty fast. Bignum can use GMP for division and base conversion if available.
`time_of{Integer.sqrt(n)} ≒ 2 * time_of{n/sqrt_n}`
`time_of{BigDecimal(Integer.sqrt(bigdecimal.to_i))} ≒ 4 * time_of{Integer.sqrt(n)}`

## Benchmark of `x.sqrt(prec)`
```
x: [2, 121e100, 2e100, '2.'+'2'*100]
prec: [8, 12, 16, 24, 32, 50, 75, 100, 150, 200, 1000, 10000]
```

| Number | Faster case | Slower case |
| --- | --- | --- |
| BigDecimal('121e100') | - |  Always slower(x12 ~ x130) |
| BigDecimal(2) | prec = 8, 24, 50, 75, 150, 200, 1000, 10000:  faster(x2 ~ x1000)  | prec = 12, 16, 32, 100:  slower(x3 ~ x4) |
| BigDecimal('2e100') | prec except 150: faster(x2 ~ x20) | prec = 150: slower(x1.3) |
| BigDecimal('2.'+'2'*100) | Always faster(x3 ~ x2000) | - |

### DIFF
--- a/lib/bigdecimal.rb
+++ b/lib/bigdecimal.rb
@@ -3,3 +3,27 @@ if RUBY_ENGINE == 'jruby'
 else
   require 'bigdecimal.so'
 end
+
+class BigDecimal
+
+  # Returns the square root of the value.
+  #
+  # Result has at least prec significant digits.
+  #
+  def sqrt(prec)
+    if infinite? == 1
+      exception_mode = BigDecimal.mode(BigDecimal::EXCEPTION_ALL)
+      raise FloatDomainError, "Computation results in 'Infinity'" if exception_mode.anybits?(BigDecimal::EXCEPTION_INFINITY)
+      return INFINITY
+    end
+    raise ArgumentError, 'negative precision' if prec < 0
+    raise FloatDomainError, 'sqrt of negative value' if self < 0
+    raise FloatDomainError, "sqrt of 'NaN'(Not a Number)" if nan?
+
+    prec = [prec, n_significant_digits].max
+    ex = prec + BigDecimal.double_fig - exponent / 2
+    base = BigDecimal(10) ** ex
+    sqrt = Integer.sqrt(self * base * base)
+    BigDecimal(sqrt).div(base, prec + BigDecimal.double_fig)
+  end
+end
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -2414,31 +2414,6 @@ BigDecimal_abs(VALUE self)
     return VpCheckGetValue(c);
 }
 
-/* call-seq:
- * sqrt(n)
- *
- * Returns the square root of the value.
- *
- * Result has at least n significant digits.
- */
-static VALUE
-BigDecimal_sqrt(VALUE self, VALUE nFig)
-{
-    ENTER(5);
-    Real *c, *a;
-    size_t mx, n;
-
-    GUARD_OBJ(a, GetVpValue(self, 1));
-    mx = a->Prec * (VpBaseFig() + 1);
-
-    n = check_int_precision(nFig);
-    n += VpDblFig() + VpBaseFig();
-    if (mx <= n) mx = n;
-    GUARD_OBJ(c, NewZeroWrapLimited(1, mx));
-    VpSqrt(c, a);
-    return VpCheckGetValue(c);
-}
-
 /* Return the integer part of the number, as a BigDecimal.
  */
 static VALUE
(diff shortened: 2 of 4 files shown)


mrkn:
> But I think that both performance gain and maintainability gain might be worth enough.

This approach’s performance advantage stems from the current, sub-optimal implementations of BigDecimal’s multiplication and division. However, once those operations are optimized, the advantage will likely disappear. Therefore, I’m willing to accept this change as a provisional improvement to the performance of sqrt.


By the way, I'm interested in why the slower cases are slower.  Do you know the reasons?

lib/bigdecimal.rb
```diff
@@ -3,3 +3,27 @@
...
+      return INFINITY
+    end
+    raise ArgumentError, 'negative precision' if prec < 0
+    raise FloatDomainError, 'sqrt of negative value' if self < 0
+    raise FloatDomainError, "sqrt of 'NaN'(Not a Number)" if nan?
+
+    prec = [prec, n_significant_digits].max
+    ex = prec + BigDecimal.double_fig - exponent / 2
+    base = BigDecimal(10) ** ex
+    sqrt = Integer.sqrt(self * base * base)
```
mrkn, line 26: How about adding the new methods for decimal shift operation to calculate the multiplication by $10^n$ faster?

tompng:
> By the way, I'm interested in why the slower cases are slower

`BigDecimal('121e100')` case is slower because the original C implementation has a fast path of convergence.
After adding a fast path 73825c8, slowness improved from `x12~x100 slower` to `x3~x10 slower`.

`BigDecimal(2)` case, the original C implementation is randomly fast or slow depend on precision.
```ruby
irb(main):005> x=BigDecimal(2); 10000.times{x.sqrt 50}
processing time: 0.336679s
irb(main):006> x=BigDecimal(2); 10000.times{x.sqrt 100} # should be slower, but 10 times faster
processing time: 0.024903s
```

lib/bigdecimal.rb
tompng: Cost of multiplication/division with 10**n is not dominant in normal path, but in the fast path such as  `BigDecimal('16e1000000').sqrt(1000000)`, the improvement of using decimal shift operation is significant.

mrkn: How much does the calculation of `10**n` cost?  I think we can further reduce the total cost by following two ways:

1. Introduce the decimal shift operation I mentioned previously.
2. Use `BigDecimal("10e#{ex}")` instead of using `BigDecimal#**` for calculating `10**n`.

We reduce the cost of `self * base * base` by the former, and the cost of calculating `base` by the latter.

Instead of the latter way, implementing the fast pass for `10**n` in `VpPowerByInt` is ok to me.

tompng: | scenario | decimal_shift | `"1e{prec}"` | `ten**prec` |
| --- | ------------ | ------------- | ----------- |
| prec=10, 100000.times | 0.149624s | 0.244520s | 0.245961s |
| prec=100, 100000.times | 0.442031s | 0.481656s | 0.605150s |
| prec=1000, 100000.times | 2.846068s | 2.963888s | 3.544875s |
| prec=10000, 10000.times | 3.441482s | 3.443580s | 3.844375s |

Comparision of:
```ruby
BigDecimal(Integer.sqrt(BigDecimal(2)._decimal_shift(2*prec)))._decimal_shift(-prec)

Integer.sqrt(BigDecimal(2)*BigDecimal("1e#{2*prec}"))*BigDecimal("1e-#{prec}")

base = BigDecimal(10)**prec; Integer.sqrt(BigDecimal(2)*base*base)/base
```

tompng: Updated the implementation c804b09e27e88bc99fb108376a6f2845d4f46539 with `* BigDecimal("1e#{2 * ex}")` and `* BigDecimal("1e#{-ex}")` (this is 4 times faster than division)

lib/bigdecimal.rb
```diff
@@ -3,3 +3,36 @@
...
+      exception_mode = BigDecimal.mode(BigDecimal::EXCEPTION_ALL)
+      raise FloatDomainError, "Computation results in 'Infinity'" if exception_mode.anybits?(BigDecimal::EXCEPTION_INFINITY)
+      return INFINITY
+    end
+    raise ArgumentError, 'negative precision' if prec < 0
+    raise FloatDomainError, 'sqrt of negative value' if self < 0
+    raise FloatDomainError, "sqrt of 'NaN'(Not a Number)" if nan?
+
+    n_digits = n_significant_digits
+    prec = [prec, n_digits].max
```
tompng, line 24: This part is re-implementing #354

tompng:
We're doing it with #381

PR #325, opened 2025-05-23
Prevent exponential precision growth in division

Confusion:
Solves Github issue #222.

Currently repeated division of a number can cause memory use to spike due to the precision of that number growing exponentially. A solution, also employed by various other arbitrary arithmetic libraries, is to only allow the precision to grow with a constant number of digits per operation.

The constant was chosen somewhat arbitrarily by me; I'm not sure what would be good reasons to use e.g. 50 or 200 or ...


mrkn:
#329 fixed the problem without introducing the precision limit, so I'm closing this issue. Thank you for your contribution!

Issue #332, opened 2025-05-30
Make releasing process a GitHub workflow


mrkn:
I used to release new versions of the gem by running the `rake release` command in my local terminal.

Since #250, we have supported JRuby through a gem targeting the Java platform. However, I always forgot that it needs to be released separately from the CRuby version.

I reflected on why I keep forgetting and identified two reasons:

1. I've been releasing many versions for a long time using a single `rake release` command.
2. I rarely use JRuby and don't have a suitable environment for it.

Defining a workflow to release both versions together would help prevent this issue.

hsbt:
I can help this same as https://github.com/ruby/psych/blob/master/.github/workflows/push_gem.yml

After that, we can publish new releases of `ruby` and `java` platforms with `git tag` and pushing that tags.

hsbt:
![Image](https://github.com/user-attachments/assets/f17b736e-0e5a-485d-b033-250bee001437)

I created the trusted publisher for `bigdecimal` now.

PR #333, opened 2025-05-30
Enabled trusted publisher for rubygems.org

hsbt:
Fixed https://github.com/ruby/bigdecimal/issues/332

I will prepare the trusted publisher of rubygems.org after merging this.


mrkn:
@hsbt Thank you!

PR #343, opened 2025-06-05
Improve sqrt calculation

tompng:
There are two pull request with different approach to improve sqrt performance.
This one improves existing C-implemented newton's method.
Another one #319 is to use implement it in Ruby using `Integer.sqrt`.

Reduce iteration count of Newton's method

Speeds up in every scenario
```ruby
# SQRT2
two = BigDecimal(2);

10000.times { two.sqrt(10) }
# processing time: 0.119187s → 0.015462s

10000.times { two.sqrt(32) }
# processing time: 0.026030s → 0.022965s

10000.times { two.sqrt(100) }
# processing time: 0.046182s → 0.029414s

10000.times { two.sqrt(1000) }
# processing time: 2.851953s → 0.623147s

two.sqrt(10000)
# processing time: 0.379927s → 0.018168s

# Fast path
BigDecimal(121).sqrt(10000)
# processing time: 0.020247s → 0.000196s

# Many digits
BigDecimal(2).div(7, 10000).sqrt(10000)
# processing time: 1.004524s → 0.017824s
# note: using Integer.sqrt is faster(0.002656s) in this case
```

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -7015,7 +7015,8 @@ VpVtoD(double *d, SIGNED_VALUE *e, Real *m)
 	goto Exit;
     }
     /* Normal number */
-    fig = roomof(BIGDECIMAL_DOUBLE_FIGURES, BASE_FIG);
+    // If frac[0] is small, one more DECDIG is needed to get enough precision.
+    fig = roomof(BIGDECIMAL_DOUBLE_FIGURES, BASE_FIG) + 1;
     ind_m = 0;
     mm = Min(fig, m->Prec);
     *d = 0.0;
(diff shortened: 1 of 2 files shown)


ext/bigdecimal/bigdecimal.c
```diff
@@ -7192,6 +7193,46 @@ VpItoV(Real *m, SIGNED_VALUE ival)
...
+        return;
+    }
+
+    // y0 = (1 ± ε) * √x
+    // y = (y0 + x/y0) / 2 ≒ (1 ± ε**2 / 2) * √x
+    // Requirements: ε**2 / 2 < 10**(-digits)
+    // ε = 10**(-(digits+1)/2) is enough
+    // Required digits for y0 should be at least (digits+1)/2+1
+    // Adding some safe margin because it does not affect performance
+    VpSqrtNewton(f, r, x, y, 3 + digits / 2);
```
tompng, line 7224: Changing this to `(digits+1) / 2 + 1` still passes this test.
```ruby
10000.times do
  x=BigDecimal("#{1+rand}e#{rand(-100..100)}")
  prec=rand(10..1000)
  raise unless (x.sqrt(prec)/x.sqrt(prec+100)-1).exponent <= -prec-10
end
```
(Initially, it failed in some case because there was a bug in VpVtoD)

Issue #340, opened 2025-06-03
TypeError/ArgumentError can't be coerced into BigDecimal or Segmentation fault after gc compaction


tompng:
Found while debugging #339, Related to #337 

`((Real*)x)->obj`, a reference from T_DATA to VALUE, will be garbled after gc compaction.
```c
typedef struct {
    VALUE  obj;     /* Back pointer(VALUE) for Ruby object.     */
    size_t MaxPrec;
    size_t Prec;
    SIGNED_VALUE exponent;
    short  sign;
    short  flag;
    DECDIG frac[FLEXIBLE_ARRAY_SIZE];
} Real;
```

After compaction, `obj` can be a reference to any kind of object.
The usage of obj field is limited. We should remove it from struct.

### Reproduction code

```ruby
p BigDecimal::VERSION
#=> "3.1.9", "3.2.1"

x = BigDecimal('1.5')
nan = BigDecimal("NaN")
inf = BigDecimal("Infinity")

GC.verify_compaction_references(expand_heap: true, toward: :empty)

x.divmod(inf) # Segfault
x.remainder(inf) # Segfault
nan + nan # Segfault
nan - nan # Segfault
BigMath.log(x, 10) # ??? can't be coerced into BigDecimal (TypeError)
BigMath.exp(x, 10) # Segfault
x**x # ??? can't be coerced into BigDecimal (ArgumentError)
x.power(x, 10) # ??? can't be coerced into BigDecimal (ArgumentError)
```

Issue #82, opened 2017-11-30
BigMath.exp is slow for large argument.


mrkn:
The original issue is https://bugs.ruby-lang.org/issues/3270

It can be optimized for the large argument by the following formula.

```
exp(a * b + c) = exp(a)**b * exp(c)
```

tompng:
Fixed by #347
```ruby
BigMath.exp(100000, 100).to_i
# 0.402142s → 0.001299s
BigMath.exp(1e19, 100)
# Infinite seconds → 0.000842s
```

PR #347, opened 2025-06-08
Implement exp, log, power and ** in ruby

### LINKED ISSUE #345: BigDecimal#** shouldn't return Integer
`BigDecimal#**` is expect to return BigDecimal

```c
/*  call-seq:
 *    self ** other -> bigdecimal
 *
 *  Returns the \BigDecimal value of +self+ raised to power +other+:
 * ...
 */
```

```ruby
BigDecimal(123).power(10, 10)
# => 0.792594609605189126649e21 (BigDecimal) OK except too high precision
BigDecimal(123).power(10.5, 10)
# => 8790299449000000000000 (Integer) Questionable result

BigDecimal(123) ** BigDecimal('10.5')
#=> 0.879029945260967009185694434e22 BigDecimal, OK
# => 8790299449000000000000 (Integer) Definitely a bug.
```

`BigMath.exp` returning Integer is the cause.
```ruby
# /* call-seq:
#  *   BigMath.exp(decimal, numeric)    -> BigDecimal

BigMath.exp(10, 5)
# => 22026 (Integer)
BigMath.exp(10, 6)
# => 0.220265e5 (BigDecimal)
```

tompng:
Writing in Ruby is easier than C.
This will bring both maintainability and performance (because implementing faster algorithm is easy).
Eliminates problem like #340 in the long run.
Use case of BigMath.log BigMath.exp are high-precision calculation (if not, using Float is enough).
Cost of multiplications/divisions are dominant. There is not much need to implement it in C.

Fixes #345
Fixes power, log and exp part of #340
Makes BigMath.exp(large_x) faster (#82)
Makes `BigMath.log(10)` which is the bottleneck of `BigDecimal#**` faster


lib/bigdecimal.rb
```diff
@@ -3,3 +3,263 @@
...
+
+    # log(x) = log(sqrt(sqrt(sqrt(sqrt(x))))) * 2**sqrt_steps
+    sqrt_steps = [2 * Integer.sqrt(prec) + 3 * x_minus_one_exponent, 0].max
+
+    # Reduce sqrt_step until sqrt gets fast
+    # https://github.com/ruby/bigdecimal/pull/323
+    # https://github.com/ruby/bigdecimal/pull/343
+    sqrt_steps /= 10
+
+    prec2 = prec + [-x_minus_one_exponent, 0].max + (sqrt_steps * 0.3010299956639812).ceil
```
tompng, line 196: `0.3010299956639812 == Math.log10(2)`
`x = sqrt(x)` makes `x-1` about half. (`sqrt(1+delta) ≒ 1+delta/2`)
performing sqrt 10 times makes `x-1` about 1/1024.
`sqrt(sqrt(sqrt(sqrt(sqrt(sqrt(sqrt(sqrt(sqrt(sqrt(1.0001)))))))))) ≒ 1.0000001`
The calculation result requires `log10(2**10)` extra digits.

lib/bigdecimal.rb
```diff
@@ -3,3 +3,263 @@
...
+        exponent += 1
+      end
+      return log10 * exponent + log(x, prec)
+    end
+
+    x_minus_one_exponent = (x - 1).exponent
+    prec += BigDecimal.double_fig
+
+    # log(x) = log(sqrt(sqrt(sqrt(sqrt(x))))) * 2**sqrt_steps
+    sqrt_steps = [2 * Integer.sqrt(prec) + 3 * x_minus_one_exponent, 0].max
```
tompng, line 189: To minimize the computation cost, we need to minimize this:
```
sqrt_cost * sqrt_steps + mult_cost * taylor_steps
```

Taylor steps can be approximated as:
```
taylor_steps = const1 * prec / (-log10(|x-1| / 2**sqrt_steps))
```

Assuming sqrt_cost and mult_cost are ideally the same order, (note that currently it is not)
We need to minimize something like this:
```
sqrt_steps + const1 * prec / (log10(|x-1|) + log10(2) * sqrt_step)
```
sqrt_steps that minimizes the cost should be
```
sqrt_steps = const2 * sqrt(prec) + log10(|x-1|) / log10(2)
           ≒ const2 * sqrt(prec) + 3 * x_minus_one_exponent
```

PR #349, opened 2025-06-10
Fix edgecase segfault of BigDecimal#remainder

tompng:
There is a possibility where coerced remainder returns `Qnil`.
`NIL_P(BigDecimal_divremain(self, r, &d, &rv))` doesn't mean `d` and `rv` has a calculated value.

This pull request is kind of a refactoring, but also fixes this segmentation fault.
```ruby
obj = Object.new
def obj.coerce(*)=[self, self]
def obj.remainder(*) = nil
BigDecimal(1).remainder(obj) #=>  [BUG] Segmentation fault at 0x0000000000000020
```


ext/bigdecimal/bigdecimal.c
```diff
@@ -2043,14 +2043,14 @@ BigDecimal_divremain(VALUE self, VALUE r, Real **dv, Real **rv)
 	b = GetVpValue(r, 0);
     }
 
-    if (!b) return DoSomeOne(self, r, rb_intern("remainder"));
+    if (!b) return Qfalse;
```
tompng, line 2046: I don't think using `Qfalse` and `Qtrue` instead of `false` and `true`, but this pattern is frequently used in `bigdecimal.c` even if the function name starts with `BigDecimal_`.
But this pattern is used in `is_zero`, `is_one` and `BigDecimal_DoDivmod`.
I think it's better to refactor in a separate pull request

PR #358, opened 2025-06-26
Refine SAVE and GUARD_OBJ macros

nobu:
Get rid of risk of `vStack` overflow.
As a trade off:
- the argument of `SAVE` must be a variable
- multiple `GUARD_OBJ` cannot be placed at one line

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -69,14 +69,14 @@ static struct {
 } rbd_rounding_modes[RBD_NUM_ROUNDING_MODES];
 
 /* MACRO's to guard objects from GC by keeping them in stack */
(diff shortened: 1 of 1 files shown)


tompng:
I'm planning to remove `obj` field from `struct Real` in #344.
This back pointer is the main reason why GUARD_OBJ is needed.
```c
typedef struct {
    VALUE  obj;     /* Back pointer(VALUE) for Ruby object.   */
    size_t MaxPrec;
    size_t Prec;
    SIGNED_VALUE exponent;
    short  sign;
    short  flag;
    DECDIG frac[FLEXIBLE_ARRAY_SIZE];
} Real;
```

After #344, most GUARD_OBJ can be removed, and we only need few RB_GC_GUARD, I think.

PR #360, opened 2025-06-30
Update docs for to_d core extensions

dduugg:
Updated docs to include Complex#to_d and NilClass#to_d, with examples.


tompng (approved):
:+1:

PR #381, opened 2025-07-18
Rewrite BigDecimal#sqrt in ruby with improved Newton's method

tompng:
Implement sqrt in ruby with faster Newton's method implementation.

Related to #323 (doing the same rewrite with Integer.sqrt)

Example calculating BigDecimal(2).sqrt in 116 precision
```ruby
x = BigDecimal(2)
y = BigDecimal(Math.sqrt(x.to_f))
y = (y+x.div(y, 18)).div(2, 18)
y = (y+x.div(y, 32)).div(2, 32)
y = (y+x.div(y, 60)).div(2, 60)
y = (y+x.div(y, 116)).div(2, 116)
```
Minimum loop count, minimum precision in each step.
The original c implementation was slow for large precision because it was not doing this.

Small precision: gets slower
Large precision: faster
```
# SQRT2
two = BigDecimal(2);

10000.times { two.sqrt(10) }
# processing time: 0.007694s → 0.073325s

10000.times { two.sqrt(32) }
# processing time: 0.020750s → 0.091189s

10000.times { two.sqrt(100) }
# processing time: 0.049758s → 0.113781s

10000.times { two.sqrt(1000) }
# processing time: 2.865216s → 0.777328s

two.sqrt(10000)
# processing time: 0.995833s → 0.017588s

# Fast path
BigDecimal(121).sqrt(10000)
# processing time: 0.000140s → 0.000154s

# Many digits
BigDecimal(2).div(7, 10000).sqrt(10000)
# processing time: 0.980485s → 0.014949s
# note: using Integer.sqrt is faster(0.002656s) in this case
```

### DIFF
--- a/lib/bigdecimal.rb
+++ b/lib/bigdecimal.rb
@@ -189,6 +189,8 @@ class BigDecimal
     raise FloatDomainError, "sqrt of 'NaN'(Not a Number)" if nan?
     return self if zero?
 
+    limit = BigDecimal.limit.nonzero? if prec == 0
+
     # BigDecimal#sqrt calculates at least n_significant_digits precision.
     # This feature maybe problematic for some cases.
     n_digits = n_significant_digits
@@ -202,6 +204,7 @@ class BigDecimal
     precs.reverse_each do |p|
       y = y.add(x.div(y, p), p).div(2, p)
     end
+    y = y.mult(1, limit) if limit
     y.mult(BigDecimal("1e#{ex}"), precs.first)
   end
 end
--- a/test/bigdecimal/test_bigdecimal.rb
+++ b/test/bigdecimal/test_bigdecimal.rb
@@ -2459,6 +2459,7 @@ class TestBigDecimal < Test::Unit::TestCase
     x = BigDecimal(123).div(7, prec)
     y = BigDecimal(456).div(11, prec)
     sqrt = BigMath.sqrt(x, prec)
+    sqrt_lim = sqrt.mult(1, limit)
     exp = BigMath.exp(x, prec)
     log = BigMath.log(x, prec)
     pow = x.power(y, prec)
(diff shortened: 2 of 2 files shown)


lib/bigdecimal.rb
```diff
@@ -3,3 +3,37 @@
...
+class BigDecimal
+
+  # Returns the square root of the value.
+  #
+  # Result has at least prec significant digits.
+  #
+  def sqrt(prec)
+    if infinite? == 1
+      exception_mode = BigDecimal.mode(BigDecimal::EXCEPTION_ALL)
+      raise FloatDomainError, "Computation results in 'Infinity'" if exception_mode.anybits?(BigDecimal::EXCEPTION_INFINITY)
```
mrkn, line 16: @tompng, how do you think about extracting the infinity check and exception-raising part as a method?

mrkn (commented):
@tompng I like this approach. Please check a comment I left.

lib/bigdecimal.rb
```diff
@@ -3,3 +3,37 @@
...
+    raise FloatDomainError, "sqrt of 'NaN'(Not a Number)" if nan?
+    return self if zero?
+
+    # BigDecimal#sqrt calculates at least n_significant_digits precision.
+    # This feature maybe problematic for some cases.
+    n_digits = n_significant_digits
+    prec = [prec, n_digits].max
+
+    ex = exponent / 2
+    x = self * BigDecimal("1e#{-ex * 2}")
```
mrkn, line 30: @tompng, I think this square root calculation can be faster if we use decimal shift operations here and on line 37, as we previously discussed. What about this?

lib/bigdecimal.rb
tompng: This is sqrt-specific condition. For example, `Math.atan(infinity)=π/2`, `"Computation results in 'Infinity'" is not a correct message.

We can re-consider adding a common infinity check method later, for unbounded monotonic increasing function and other types, but I think it's too early for now.

PR #398, opened 2025-08-09
Faster log(10) calculation

tompng:
Improve performance of `log(10)` by using `log(10)=8*log(4/3)+log(32805/32768)`.
Improving log(10) will improve performance of log(x) of x<=0.1 and x>=10.

```ruby
BigMath.log(10, 10000)
# master branch:            15.398019s
# pull #381 (sqrt improve):  1.313900s
# this pull request:         0.167481s
```


tompng:
Diff is not so simple, and not so optimal.
```ruby
BigMath.log(BigDecimal(2), 250000)
# This optimization: 86.393285 seconds
# sqrt(pull #381) + mult(pull #380) optimization: 77.847769 seconds
```
I'll revisit this after introducing `BigMath.log2` and `BigMath.log10`

PR #417, opened 2025-09-02
In JRuby, don't add sqrt, exp, log, power implemented in ruby

tompng:
Stub out ruby-implemented methods (`sqrt` `exp` `log` `power`) in JRuby for now.
These methods already exist in JRuby.
It depends on new feature only implemented in bigdecimal.c: `BigDecimal(float_without_prec)` and `BigDecimal#_decimal_shift`

There are some utility functions in `lib/bigdecimal/bigdecimal.rb`: `BigDecimal::Internal.coerce_to_bigdecimal`, `BigDecimal::Internal.validate_prec`, etc.
I want to use them in BigMath in the future. At that time, I think we need to remove this `return` and fix the ruby code.

### DIFF
--- a/lib/bigdecimal.rb
+++ b/lib/bigdecimal.rb
@@ -1,5 +1,6 @@
 if RUBY_ENGINE == 'jruby'
   JRuby::Util.load_ext("org.jruby.ext.bigdecimal.BigDecimalLibrary")
+  return
 else
   require 'bigdecimal.so'
 end


tompng:
Thank you :+1:
Since `_decimal_shift` is currently an internal API (name may change), I'd like to implement _decimal_shift in ruby/bigdecimal using `move_point_left` and `move_point_right` soon.
But doing it, I think we need a CI for JRuby for at least methods defined in `bigdecimal/bigdecimal.rb` and `bigdecimal/math.rb`.

PR #436, opened 2025-10-02
Test memory_leak

nobu:


tompng:
Looks like test is passing in master now, and memory-leak test seems correctly running on ci. Thank you!

CI log, failing correct on a branch that adds memory leak bug: https://github.com/tompng/bigdecimal/actions/runs/18289780063/job/52073894712

PR #528, opened 2026-04-15
Update dtoa to version from Ruby 4.0

### LINKED ISSUE #527: dtoa is not ractor safe
This code crashes.
```ruby
4.times.map do
  Ractor.new do
    loop { raise unless BigDecimal(3.14) }
  end
end
sleep 1
GC.start
exit
```

Simplifying `rb_float_convert_to_BigDecimal` to always return `BigDecimal(1)`, I discovered that `BigDecimal_dtoa` is suspicious.
The above ruby code only crash if the `dtoa` part is uncommented in the code below.
```c
static VALUE
rb_float_convert_to_BigDecimal(VALUE val, size_t digs, int raise_exception)
{
    // int decpt, negative_p;
    // char *e;    
    // char *p = BigDecimal_dtoa(3.14, 0, 0, &decpt, &negative_p, &e);
    // xfree(p);
    return rb_inum_convert_to_BigDecimal(INT2FIX(1));
}
```

jhawthorn:
Fixes #527

Upstream commits:
  - [3acc81d9e4](https://github.com/ruby/ruby/commit/3acc81d9e4) Fixed race in dtoa [Bug #17612]
  - [d0a822eec5 ](https://github.com/ruby/ruby/commit/d0a822eec5)Fix dtoa buffer overrun
  - [bc8cc68aef](https://github.com/ruby/ruby/commit/bc8cc68aef) Make dtoa.c buildable alone
  - [d17edf3a17 ](https://github.com/ruby/ruby/commit/d17edf3a17)[Bug #20705] Update strtod implementation
  - [b3b1d7b295](https://github.com/ruby/ruby/commit/b3b1d7b295) Remove spinlock in dtoa's Balloc, use xmalloc
  - [0887d82406](https://github.com/ruby/ruby/commit/0887d82406) dtoa.c: Add shortcut if arguments are zero
  - [d0c966adcd](https://github.com/ruby/ruby/commit/d0c966adcd) dtoa.c: Extract macro to update 5powers Bigint cache atomically
  - [f047174cf1](https://github.com/ruby/ruby/commit/f047174cf1) dtoa.c: Check memory allocation failures
  - [624538ba92](https://github.com/ruby/ruby/commit/624538ba92) dtoa.c: Check integer underflow
  - [44693ee329](https://github.com/ruby/ruby/commit/44693ee329) Fix a possible memory leak in dtoa

### DIFF
--- a/ext/bigdecimal/bigdecimal.c
+++ b/ext/bigdecimal/bigdecimal.c
@@ -2709,7 +2709,7 @@ rb_float_convert_to_BigDecimal(VALUE val, size_t digs, int raise_exception)
         len10 = BIGDECIMAL_DOUBLE_FIGURES;
     }
     memcpy(buf, p, len10);
-    xfree(p);
+    free(p);
 
     VALUE inum;
     size_t RB_UNUSED_VAR(prec) = 0;
--- a/ext/bigdecimal/missing/dtoa.c
+++ b/ext/bigdecimal/missing/dtoa.c
@@ -98,17 +98,6 @@
  *	if memory is available and otherwise does something you deem
  *	appropriate.  If MALLOC is undefined, malloc will be invoked
  *	directly -- and assumed always to succeed.
- * #define Omit_Private_Memory to omit logic (added Jan. 1998) for making
- *	memory allocations from a private pool of memory when possible.
- *	When used, the private pool is PRIVATE_MEM bytes long:  2304 bytes,
- *	unless #defined to be a different length.  This default length
- *	suffices to get rid of MALLOC calls except for unusual cases,
- *	such as decimal-to-binary conversion of a very long string of
- *	digits.  The longest string dtoa can return is about 751 bytes
- *	long.  For conversions by strtod of strings of 800 digits and
- *	all dtoa conversions in single-threaded executions with 8-byte
- *	pointers, PRIVATE_MEM >= 7400 appears to suffice; with 4-byte
- *	pointers, PRIVATE_MEM >= 7112 appears adequate.
  * #define INFNAN_CHECK on IEEE systems to cause strtod to check for
  *	Infinity and NaN (case insensitively).  On some systems (e.g.,
  *	some HP systems), it may be necessary to #define NAN_WORD0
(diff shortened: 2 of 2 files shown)


ext/bigdecimal/bigdecimal.c
```diff
@@ -2709,7 +2709,7 @@ rb_float_convert_to_BigDecimal(VALUE val, size_t digs, int raise_exception)
         len10 = BIGDECIMAL_DOUBLE_FIGURES;
     }
     memcpy(buf, p, len10);
-    xfree(p);
+    free(p);
```
byroot, line 2712: Shouldn't we fix `dtoa` to use `xmalloc` instead?

jhawthorn: I'm not sure there's value in doing so. These allocations are short-lived and running the GC doesn't help us remove them (they aren't associated with a Ruby heap object) so I'm not sure it helps for them to contribute to GC pressure.

This was changed upstream in https://github.com/ruby/ruby/commit/bc8cc68aeff7ae3be1353f4077633cda4d30b54b

byroot: Sorry, I should have explained my train of thoughts. I agree there's no GC benefits, especially since when you use `xfree` (instead of the new `xfree_sized`) this likely cause a `malloc_usable_size` call which make the whole thing slower.

However, I recently ran into a bunch of bugs related to using `xfree` with `malloc` and vice-versa, so I tend to think it's simpler to just always use the Ruby allocator.

But really up to you.

Anyway, as you said `bgidecimal` should probably move away from vendoring `dtoa`.

tompng: I'd like to use `malloc` and `free` instead of `xmalloc` and `xfree` in bigdecimal's dtoa.
`dtoa` and `strtod` may call `MALLOC` several times, and if the second call fails to allocate, dtoa frees the first allocated one.
```c
b1 = Balloc(b->k+1);
if (!b1) {
    Bfree(b);
    return NULL;
}
```

```c
bd = Balloc(bd0->k);
if (!bd) goto retfree;
```
If we use ruby_xmalloc, and the second allocation raises NoMemoryError, this mechanism will fail, and will cause memory leak.
So I think current `dtoa.c` in ruby/ruby can be considered designed to be use with `malloc` and `free`

>  move away from vendoring `dtoa`.

Agree

tompng (approved):
Thank you!

PR #548, opened 2026-08-29
Keep Newton division divisor alive

OskarEichler:
Summary:
- correct the second duplicated RB_GC_GUARD for a2 to guard b2
- retain the divisor wrapper through the end of VpDivdNewtonInner

Why:
The method constructs a2, b2, c2 and r2, but its final guards name a2 twice and omit b2. The duplicate is a maintenance/liveness typo at a native GC boundary.

Verification:
- 25 GC.stress division comparisons pass before and after the correction

Compatibility:
No arithmetic behavior changes. This makes the existing lifetime intent explicit; the stress model did not reproduce a user-visible crash on the tested runtime.


tompng (approved):
Nice catch! Thanks

PR #550, opened 2026-08-29
Honor exception: false for strings with null bytes

OskarEichler:
Summary:
- detect embedded NUL bytes before StringValueCStr in non-raising conversion mode
- return nil when exception: false
- retain the existing ArgumentError when exceptions are enabled

Why:
StringValueCStr raises string contains null byte before BigDecimal’s parser can honor exception: false. Core Float and Integer conversions use their non-raising option for this case, and BigDecimal already documents the same conversion mode.

Verification:
- baseline reproduction raises for BigDecimal("1\0", exception: false)
- fixed reproduction returns nil; exception: true still raises; ordinary strings are unchanged

Compatibility:
Only strings invalid because of an embedded NUL change in the explicitly non-raising mode.


tompng:
Looks good. Could you add a test for this?

tompng (approved):

